PR #3, opened 2026-01-03
Fix transaction matching and add block catch-up processing

shruggr:
### Block Processing & Transaction Matching
- Fix critical bug in `fetchHashes`: use `chainhash.NewHash(bytes)` instead of roundtrip through hex encoding which caused transaction matching to fail
- Add block tracking with `processed_blocks` table to track on_chain status
- Subscribe to chaintracks tip updates for block catch-up processing
- Add `fetchBlockSubtrees` to properly parse teranode binary block format
- Handle reorgs by resetting transaction statuses when blocks are orphaned
- Add enhanced logging for block processing and transaction matching
- Add ARC error types for proper HTTP status codes on validation failures

### Standalone Server & Documentation
- Remove chaintracks routes from standalone arcade server (decouple for standalone publishing)
- Add WebhookHandler to embedded services initialization
- Add `combined_server` example showing how to run arcade+chaintracks together
- Fix README documentation (API paths, config options, endpoints, component descriptions)
- Fix SSE client examples (correct URL format and port)
- Fix Swagger annotations and regenerate docs
- Add debug logging to webhook handler

## Test plan

- [x] Verified transaction `5d8325f4535869f3902e2c93749710214c2ccccc0ad70483bd34f196574e110b` is now correctly matched and marked as MINED
- [x] Verified block catch-up processing works after clearing processed blocks
- [x] Verified all standalone endpoints work correctly (`/tx`, `/txs`, `/policy`, `/events`, `/health`)
- [x] Verified OpenAPI spec has correct paths
- [x] Build passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,21 +43,10 @@ storage_path: ~/.arcade
 server:
   address: ":3011"
 
-database:
-  type: sqlite
-  sqlite_path: ~/.arcade/arcade.db
-
-events:
-  type: memory
-  buffer_size: 1000
-
 teranode:
-  # Set via ARCADE_TERANODE_BASE_URL environment variable
+  broadcast_urls:
+    - "https://arc.taal.com"
   timeout: 30s
-
-validator:
-  max_tx_size: 4294967296
-  min_fee_per_kb: 100
 ```
 
 ### Run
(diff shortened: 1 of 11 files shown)


sirdeggen (approved):
Let's just get it in, remaining tweaks are now issues to resolve in due course.

PR #45, opened 2026-04-07
fix: broadcast context and status code improvements

shruggr:
## Summary

- Detach broadcast goroutines from request context using `context.WithoutCancel` so client disconnect doesn't cancel in-flight teranode submissions
- Use a shared broadcast context with 15s timeout per fan-out, cleaned up when all goroutines complete (not on early return)
- Set `StatusCode` on `TransactionStatus` at construction in `submitToTeranodeSync`: pass through teranode HTTP codes, 503 for network failures, 504 for timeouts
- Simplify collection loop from `select`-based to `range` over buffered channel

## Test plan

- [ ] Verify broadcasts propagate to all endpoints even when caller returns early on first success
- [ ] Verify client disconnect mid-broadcast doesn't cancel remaining submissions
- [ ] Verify `StatusCode` is non-zero in SSE events and webhook callbacks
- [ ] Verify 503 for network failures, 504 for timeouts, passthrough for teranode 4xx/5xx
- [x] Existing tests pass, linter clean

### DIFF
--- a/service/embedded/embedded.go
+++ b/service/embedded/embedded.go
@@ -211,8 +211,7 @@ func (e *Embedded) SubmitTransaction(ctx context.Context, rawTx []byte, opts *mo
 		slog.Duration("timeout", 15*time.Second),
 	)
 	resultCh := make(chan *models.TransactionStatus, len(endpoints))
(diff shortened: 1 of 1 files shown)


mrz1836 (approved):
LGTM

PR #110, opened 2026-05-01
[Do not merge] fix(tx_validator): actually validate fees and scripts (#80)

### LINKED ISSUE #80: [F-022] tx-validator always skips fee and script validation
## Summary
tx-validator always skips fee and script validation

## Location
`services/tx_validator/validator.go:475-480`

## Severity / Category / Confidence
- Severity: HIGH
- Category: Correctness
- Confidence: High

## Code
```go
	for _, vt := range live {
		g.Go(func() error {
			if err := v.txValidator.ValidateTransaction(gctx, vt.parsed, true, true); err != nil {
				vt.rejected = true
				vt.rejectReason = err.Error()
				v.logger.Info("transaction validation failed",
```

## Issue
The service calls `ValidateTransaction` with both `skipFees` and `skipScripts` hard-coded to `true`. That bypasses the validator's fee model and script/SPV checks for every submitted transaction.

## Impact
Transactions with invalid scripts or insufficient fees can pass the local validation pipeline and be propagated, shifting rejection to downstream Teranode peers and making Arcade's validation status misleading.

## Recommendation
Pass skip flags from authenticated/request-level policy or configuration, default them to false, and only bypass checks when explicitly requested and authorized.

---
*Filed from the 2026-05 third-party code review (finding `F-022`). See `REVIEW_REPORT.md` in the review bundle for full context and cross-references.*

galt-tr:
## Summary
- The third and fourth args to `validator.ValidateTransaction` are `skipFees` and `skipScripts`. The tx_validator service was hard-coding both to `true`, which silently disabled fee and script enforcement for every submitted transaction (finding F-022).
- Plumbed two new opt-in escape hatches through `TxValidatorConfig` — `skip_fee_validation` and `skip_script_validation`, both defaulting to `false`. Production behavior is now "validate everything" by default; operators that can't currently satisfy the underlying SDK's prerequisites (e.g. no chaintracker wired up — `cmd/arcade/main.go` still constructs `validator.NewValidator(nil, nil)`) can flip the knobs explicitly rather than relying on a silently broken default.
- Added regression tests: a zero-fee tx is now rejected by default, the same tx passes when both skip flags are set, and a tiny invariant test pins the default-false config so a future refactor can't quietly turn enforcement back off.

Closes #80

## Reviewer please double-check
This change WILL cause previously-passing transactions to be rejected if production has been silently relying on the broken behavior. Two specific concerns to validate before merge:

1. **No chaintracker is currently wired up.** `cmd/arcade/main.go:133` calls `validator.NewValidator(nil, nil)`. With `skipScripts=false` the SDK's `spv.Verify` falls back to `chaintracker.NewWhatsOnChain(MainNet, "")` (see go-sdk/spv/verify.go), which means every script verification reaches out to WhatsOnChain. That is likely undesirable for arcade's deployment topology. Two reasonable mitigations the reviewer should pick from:
   - Wire arcade's existing chaintracks client into `NewValidator` here, OR
   - Default `skip_script_validation: true` in `setDefaults()` until the chaintracker is wired (and then drop the default once it is).
2. **Fee policy is the SDK default of 100 sat/kB.** Any transaction that paid <100 sat/kB before will now be rejected. Worth a sanity check against historical traffic / a staging environment before promoting to mainnet.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./services/tx_validator/... ./validator/... -count=1 -race`
- [x] `golangci-lint run ./services/tx_validator/... ./validator/...` — 0 issues
- [ ] Reviewer to verify the production behavior change is intended and won't reject anything legitimate

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -131,6 +131,15 @@ tx_validator:
 # window. Validation is CPU-bound and stateless, so the sweet spot is around
 # NumCPU. Set 0 to take runtime.NumCPU() at startup.
 # parallelism: 0
+#
+# skip_fee_validation / skip_script_validation default to false. Leave them
+# false in production — the validator will then enforce the policy minimum
+# fee and run full script verification (via spv.Verify) for every submitted
+# transaction. Set either true ONLY in tests or in environments where the
+# underlying preconditions can't be satisfied (e.g. no chaintracker is wired
+# up yet); flipping them on is equivalent to "trust whatever the client sent".
+# skip_fee_validation: false
+# skip_script_validation: false
 
 bump_builder:
   # Grace window after BLOCK_PROCESSED before reading STUMPs, to allow for
--- a/config/config.go
+++ b/config/config.go
@@ -356,8 +356,19 @@ type CallbackConfig struct {
 // single flush window — bounded so a huge in-flight batch can't open more
 // goroutines than the host has cores. Zero or negative falls back to
 // runtime.NumCPU at construction time.
+//
+// SkipFeeValidation / SkipScriptValidation default to false (i.e. fees AND
+// scripts are checked). They exist as escape hatches for environments that
+// can't currently fulfil the underlying validator's prerequisites — for
+// example, deployments without a wired-up chaintracker should set
+// skip_script_validation=true rather than letting the SDK silently fall back
+// to WhatsOnChain. Operators flipping these to true accept that fee/script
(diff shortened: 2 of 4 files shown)


galt-tr:
I don't know if we want this right now; I think we put this on ice. Right now we simply defer most script validation to teranode.

PR #123, opened 2026-05-01
fix(bump_builder): preserve block height when publishing mined status (#87)

### LINKED ISSUE #87: [F-029] mined status updates drop the block height before publishing
## Summary
mined status updates drop the block height before publishing

## Location
`services/bump_builder/builder.go:206-214`

## Severity / Category / Confidence
- Severity: HIGH
- Category: Correctness
- Confidence: High

## Code
```go
	// 5. Store compound BUMP as binary
	if err := b.store.InsertBUMP(ctx, blockHash, blockHeight, bumpBytes); err != nil {
		return fmt.Errorf("storing BUMP: %w", err)
	}

	// 6. Set tracked transactions to MINED
	if len(txids) > 0 {
		mined, err := b.store.SetMinedByTxIDs(ctx, blockHash, txids)
```

## Issue
The builder extracts `blockHeight` from the compound BUMP and persists it with the BUMP, but the mined-status call only receives `blockHash` and `txids`. All current store implementations set `Status=MINED` without writing `block_height`, and the status events published from `mined` therefore omit the height.

## Impact
API/SSE clients receive mined updates without the documented block height, and `TxTracker.LoadFromStore` cannot prune old mined rows because it only prunes when `BlockHeight > 0`.

## Recommendation
Extend `SetMinedByTxIDs` to accept the block height or have each backend read it from the stored BUMP, then persist and return `BlockHeight` in the mined statuses.

---
*Filed from the 2026-05 third-party code review (finding `F-029`). See `REVIEW_REPORT.md` in the review bundle for full context and cross-references.*

galt-tr:
## Summary
- `Store.SetMinedByTxIDs` now takes `blockHeight uint64`; postgres / aerospike / pebble all persist the height alongside `block_hash` and echo it back on each returned `TransactionStatus`.
- `bump_builder.Builder` threads `blockHeight` (lifted from the compound BUMP) into `SetMinedByTxIDs` and adds a defensive top-up on the publish path so a partial regression in any backend cannot reintroduce a zero-height MINED update.
- Added three regression tests in `services/bump_builder/builder_test.go` (a `recordingPublisher` mock, an explicit non-trivial height case via `transaction.NewMerklePath`, a "height never zero" guard, and a "publish path repairs a buggy backend" test).

Closes #87 (F-029).

### Where the height was being lost
`services/bump_builder/builder.go` captured `blockHeight` for `InsertBUMP` then called `b.store.SetMinedByTxIDs(ctx, blockHash, txids)` — no height. Every backend implementation of `SetMinedByTxIDs` therefore wrote only `block_hash` and built `models.TransactionStatus` with `BlockHeight` zero-valued, which the publisher then fanned out to SSE / webhook subscribers verbatim.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...` (and `go vet -tags=postgres ./...`, `-tags=integration`, `-tags=aerospike` clean)
- [x] `go test ./services/bump_builder/... ./store/... -count=1 -race`
- [x] `golangci-lint run ./services/bump_builder/... ./store/...` — 0 issues
- [ ] Reviewer to confirm downstream consumers benefit from the corrected height

### DIFF
--- a/services/api_server/handlers_test.go
+++ b/services/api_server/handlers_test.go
@@ -398,8 +398,7 @@ func TestHandleCallback_UnknownTxid_NoPhantomRow(t *testing.T) {
 			}
 			body := mustMarshalJSON(t, payload)
 
-			req := httptest.NewRequestWithContext(t.Context(), http.MethodPost, "/api/v1/merkle-service/callback", bytes.NewReader(body))
-			req.Header.Set("Content-Type", "application/json")
+			req := authedCallbackRequest(t, body)
 			w := httptest.NewRecorder()
 
 			router.ServeHTTP(w, req)
--- a/services/api_server/handlers_test.go
+++ b/services/api_server/handlers_test.go
@@ -78,7 +78,7 @@ func (m *mockStore) SetStatusByBlockHash(context.Context, string, models.Status)
 }
 func (m *mockStore) InsertBUMP(context.Context, string, uint64, []byte) error { return nil }
 func (m *mockStore) GetBUMP(context.Context, string) (uint64, []byte, error)  { return 0, nil, nil }
-func (m *mockStore) SetMinedByTxIDs(context.Context, string, []string) ([]*models.TransactionStatus, error) {
+func (m *mockStore) SetMinedByTxIDs(context.Context, string, uint64, []string) ([]*models.TransactionStatus, error) {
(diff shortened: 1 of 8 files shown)


mrz1836 (approved):
LGTM

PR #151, opened 2026-05-14
Plan: dependency-aware dispatch

shruggr:
Plan doc for discussion before implementation. Located at `docs/plans/dependency-aware-dispatch.md`.

## What it covers

- Tracking parent-child relationships between in-flight transactions so children are not broadcast until their parents have reached a terminal state, eliminating the cross-batch "missing inputs" race.
- Pipeline simplifications that fall out of the change:
  - Validation moves into the intake handler; the `tx_validator` service is removed
  - Propagation topic becomes single-partition with a single-goroutine dispatcher owning the dep index and retry queue
  - `PENDING_RETRY` status and the reaper are removed; retry state lives in dispatcher memory, durable via Kafka replay
- Retry classification driven by Teranode HTTP status codes (per the audit in bsv-blockchain/teranode#870), not by string-matching error bodies.

## Deployment shape

Single coordinated deploy with a new propagation topic. Operational prerequisite is draining the old queue before cutover. No backward-compat code path is retained — given the limited install base, the cost of a clean break is minimal.

## Open for discussion

The plan is intentionally scoped to the dep-aware dispatch work plus the simplifications that come with it. Out of scope: binary message format (deferred until measurement justifies it), stale `SENT_TO_NETWORK` handling (separate concern, status is unused in today's code).


services/propagation/propagator.go
```diff
@@ -111,9 +119,49 @@ type broadcastJob struct {
...
+	// terminal — dispatcher keeps the offset alive and processBatch
+	// sends the original propagation message back to the dispatcher
+	// after a short flat wait.
+	txResultClassRequeue
+	// txResultClassInFlight: at least one peer 2xx'd the broadcast but
+	// no peer gave a verdict (typically all 202). The tx is in-flight
+	// on the network and merkle-service's eventual callback flips its
+	// status. processBatch writes nothing, requeues nothing, and the
+	// dispatcher's offset stays alive until the callback handler sends
+	// a terminal event.
```
shruggr: Does this make sense for 202?

I wonder if we shouldn't treat it as processed for the sake of kafka?

Maybe we should keep reaper.go to rebroadcast things which never achieve ACCEPTED/REJECTED/MINED status?

galt-tr:
Looking good; lets address copilot feedback

galt-tr (approved):

PR #158, opened 2026-05-20
store/aerospike: keep bin names within Aerospike's 15-char limit

mboyd1:
The Aerospike store backend wrote three bins whose names exceed Aerospike's 15-character limit:

  - InsertSubmission        full_status_updates   (19)
  - UpdateDeliveryStatus    last_delivered_status (21)
  - status / MarkMerkleRegisteredByTxIDs  merkle_registered_at (20)

Aerospike rejects the entire write with BIN_NAME_TOO_LONG when any bin name is too long, so submission inserts and merkle-registration updates failed outright. This was latent under the Pebble store, which has no bin-name constraint, and only surfaced once the store backend was switched to Aerospike

Rename the offending bins to full_updates, last_status and merkle_reg_at. Model field names and the JSON wire format are unchanged

   ## Fix

   Rename the offending bins to `full_updates`, `last_status`, and `merkle_reg_at`. Model field names and
   the JSON wire format are unchanged; no records carry the old bins (every such write failed), so there is
   nothing to migrate.

   ## Testing

   Verified against an Aerospike-backed deployment: before the fix the submissions set held 0 objects with
   continuous `BIN_NAME_TOO_LONG` warnings; after the fix submissions persist and the warnings stop.

### DIFF
--- a/store/aerospike/aerospike.go
+++ b/store/aerospike/aerospike.go
@@ -378,7 +378,7 @@ func (s *Store) UpdateStatus(ctx context.Context, status *models.TransactionStat
 		bins["merkle_path"] = []byte(status.MerklePath)
 	}
 	if !status.MerkleRegisteredAt.IsZero() {
-		bins["merkle_registered_at"] = status.MerkleRegisteredAt.UnixMilli()
+		bins["merkle_reg_at"] = status.MerkleRegisteredAt.UnixMilli()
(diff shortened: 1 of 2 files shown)


mrz1836 (approved):
LGTM

galt-tr (approved):

PR #159, opened 2026-05-21
propagation: treat any per-tx Teranode failure line as terminal

shruggr:
## Summary
- Drop the terminal-code allowlist in `classifyFailureLine`; any per-tx failure line returned by Teranode is now treated as terminally rejected.
- Whole-batch requeue path (5xx with no parseable body) is unchanged — that remains the only legitimate "transient infra" signal.

## Why
Teranode wraps validator failures with `NewProcessingError` at [`services/propagation/Server.go:1236`](https://github.com/bsv-blockchain/teranode/blob/main/services/propagation/Server.go#L1236), so `PROCESSING` is the catch-all "this tx is bad" code in practice. Only a few hard-fail paths (coinbase, missing inputs/outputs, duplicate inputs, size limit) emit `TX_INVALID` directly. With the prior whitelist, real validation failures arrived as `PROCESSING`, fell into the "requeue" bucket, and looped indefinitely — arcade's propagator has no max-attempt limit, so the tx would never exit the queue on its own.

## Test plan
- [x] `go test ./services/propagation/...` passes
- [x] Manual: a tx that Teranode rejects with `PROCESSING (4): ... failed to validate transaction` now flips to `REJECTED` on the next propagator pass instead of being requeued forever.

### DIFF
--- a/services/propagation/propagator.go
+++ b/services/propagation/propagator.go
@@ -816,49 +816,20 @@ type txResult struct {
 
 // classifyFailureLine maps one Teranode /txs failure line into a dispatcher
 // action bucket. The line is the full UserMessage produced by Teranode,
-// e.g. "TX_INVALID (31): [ProcessTransaction][<txid>] ...". A line in the
-// terminal-rejection bucket (TX_INVALID, UTXO_SPENT, etc.) → rejected.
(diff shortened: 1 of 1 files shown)


mrz1836 (approved):
LGTM

PR #160, opened 2026-05-22
propagation: notify dispatcher of terminalized tx so Kafka offssets release

mboyd1:
## Summary

  `applyTerminalStatuses` only notified the propagation dispatcher of transactions
  whose store status actually *transitioned*. A transaction replayed from Kafka
  after a consumer rebalance is already at its terminal status, so the dispatcher
  was never told it finished — its Kafka offset was never released, and the
  `arcade-propagation` consumer group's commit watermark pinned permanently
  (`CURRENT-OFFSET` shows `-` in `kafka-consumer-groups --describe`). Every
  restart then re-read the entire topic from offset 0.

  ## What Changed

  - **`applyTerminalStatuses`** now notifies the dispatcher for **every** tx in
    `terminalStatuses`, unconditionally. Offset release is decoupled from whether
    the store write moved the row. The SSE/webhook bulk publish keeps its
    `prev`-row filter, so lattice no-ops and reaped rows still don't emit phantom
    events.
    on `ctx`. A `processBatch`/requeue goroutine that outlives the dispatcher on
    a claim teardown abandons its send instead of blocking forever on a dispatcher
    that has already exited. A requeue dropped on teardown is logged; the tx stays
    uncommitted and the next claim replays it (at-least-once preserved).
  - New `dispatcher_test.go` (see Testing).

  ## Why It Was Necessary

  The dispatcher commits Kafka offsets up to `offsetTracker.LowestUnfinished()` —
  the lowest admitted offset not yet `Done()`'d. Every admitted tx `Add`s its
  offset; `handleTerminal` `Done`s it on terminalization.

  `applyTerminalStatuses` built its accepted/rejected txid lists from rows where
  `prev != nil && prev.Status != st.Status`, and used that **same filtered list**
  to drive the dispatcher notification. For a Kafka-replayed tx already at its
  terminal status, `BatchUpdateStatusReturning` reports a lattice no-op
  (`prev.Status == st.Status`) or a reaped row (`prev == nil`) — so the tx was
  dropped from the notification, `handleTerminal` never ran, the offset was never
  `Done`'d, and `LowestUnfinished()` pinned forever. A single stranded low offset
  freezes the commit watermark indefinitely, even in a healthy long-lived claim.
  Symptoms: the topic is re-read from 0 on every restart, already-mined txs are
  re-processed, and sarama fetch buffers grow under the re-read load.

  ## Testing Performed

  - `dispatcher_test.go` adds:
    - Pure-function offset-accounting tests — `handleAdmit`/`handleTerminal`/
      `handleRequeue`/`cascadeReject` `Add`/`Done` bookkeeping, including the
      out-of-order and requeue-keeps-pinned cases.
    - `runDispatcher` integration tests driving a `fakeClaim` that assert
      `claim.MarkMessage` (the real commit-watermark signal) fires for
      lattice-no-op and reaped-row terminals.
    - Both integration tests **fail against pre-fix code** and pass with the fix.
  - Full `services/propagation` suite and `cmd/arcade` build pass.
  - Verified on a live deployment: after deploy, the `arcade-propagation`
    commit watermark advanced off `-` and the consumer group worked through
    its full ~2.5M-message backlog in ~1 hour (≈42k offsets/min average).
  ## Impact / Risk

  - **Behavior change:** the dispatcher now receives a terminal notification for
    every terminalized tx, not just store-transitioned ones. This is the intended
    fix — offsets get released as designed.
  - **At-least-once delivery is preserved.** Offsets are still only committed
    below `LowestUnfinished()`; a tx dropped on claim teardown stays uncommitted
    and is replayed.
  - No schema, config, or API changes. Risk is low and contained to
    `services/propagation`.
  - On first deploy, instances whose `arcade-propagation` group never committed
    will re-read the topic once more from the last (possibly absent) commit, then
    begin committing normally — a one-time catch-up, not a regression.

### DIFF
--- a/services/propagation/dispatcher.go
+++ b/services/propagation/dispatcher.go
@@ -582,10 +582,29 @@ func (p *Propagator) admitToDispatcher(msg propagationMsg, offset int64) admitRe
 // notifyTerminalToDispatcher is the post-broadcast helper. Tells the
 // dispatcher the txid reached terminal status, returns the cascaded
 // descendants (caller writes REJECTED rows for them).
-func (p *Propagator) notifyTerminalToDispatcher(txid string, status models.Status) terminalResult {
+//
+// Both the send and the reply receive are guarded by ctx. processBatch
+// goroutines outlive the dispatcher on a claim teardown (rebalance /
+// shutdown): the dispatcher loop returns as soon as its ctx is canceled,
+// so an unguarded reply receive here would block this goroutine forever
+// on a dispatcher that will never answer. On ctx cancellation the notify
+// is abandoned and a zero terminalResult is returned — the tx's Kafka
+// offset is simply left un-Done and uncommitted, and the next claim
+// replays it (at-least-once). reply is buffered so a late dispatcher
+// answer after we've stopped waiting never blocks the dispatcher.
+func (p *Propagator) notifyTerminalToDispatcher(ctx context.Context, txid string, status models.Status) terminalResult {
 	reply := make(chan terminalResult, 1)
-	p.terminalCh <- terminalEvent{txid: txid, status: status, reply: reply}
-	return <-reply
+	select {
+	case p.terminalCh <- terminalEvent{txid: txid, status: status, reply: reply}:
+	case <-ctx.Done():
+		return terminalResult{}
+	}
+	select {
+	case r := <-reply:
+		return r
+	case <-ctx.Done():
+		return terminalResult{}
+	}
 }
 
 // drainPending asks the dispatcher for the current pendingMsgs as a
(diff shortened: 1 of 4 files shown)


mrz1836 (changes requested):
LGTM - CI checks are failing.

mboyd1:
Thanks for the review.

  **Copilot review comments** — both addressed in 4d47ca4:

  - `propagator.go`: dispatcher notification is now skipped entirely when
    `BatchUpdateStatusReturning` returns an error, so a failed status write
    can't advance the Kafka commit watermark past an unpersisted tx
    (at-least-once preserved). Added
    `TestApplyTerminalStatuses_StoreError_DoesNotReleaseOffset` as a
    regression guard.
  - `dispatcher.go`: the `ctx.Done()` reply-wait now does a final
    non-blocking read of `reply`, so a claim teardown racing a ready
    dispatcher reply can't drop the cascaded txids the caller must persist.

  **CI lint failure** — fixed in 2a98f1d: the `containedctx` hit on the
  `fakeClaim` test struct (it has to hold a `context.Context` to satisfy
  `kafka.Claim.Context()`) is suppressed with a `//nolint:containedctx`
  directive, matching the existing `broadcastJob` precedent in this package.

  Branch is also synced with latest `main`. The new CI run is currently
  awaiting workflow approval (fork-PR gate) — once it runs, lint should be
  green. Ready for another look.

galt-tr (approved):

PR #162, opened 2026-05-23
perf(bump): rebuild compound BUMP directly from STUMPs; fix OOM

mboyd1:
`BuildCompoundBUMP` is arcade's hot path for assembling a block's compound merkle proof from the
  per-subtree STUMPs that merkle-service delivers. The previous implementation called
  `assembleFullPath` once per STUMP and merged the results — quadratic in subtree count, fine at a few
   thousand subtrees, fatal at Teranode's larger blocks. This PR rebuilds the compound directly from
  the same inputs in a single top-down pass.

  **Production proof, 2026-05-22, real mainnet block 950028 (24,896 STUMPs / ~199k txs):** the old
  algorithm drove arcade to **13 GB RAM and 100% CPU for >8 minutes with no completion**, on a
  trajectory to OOM-restart. Live blocks 950275 (112k txs) and 950276 (201k txs)
  would have triggered the same path had they contained arcade-tracked txs.  We built an out-of-scope 
  backlog drain tool and worked around it offline; this PR fixes the live path.

  ## What Changed

  - `bump/BuildCompoundBUMP` body rewritten in place — **same signature, same callers, same output**,
  new algorithm:
    - Each STUMP contributes its own elements once, with offsets shifted from local-to-subtree to
    - The subtree-root layer is seeded once with `subtreeHashes` for slots that have no STUMP; slots
  that do have one get their subtree root computed up from level 0 in the same top-down pass.
    - Tracked-tx `Txid` markers on STUMP leaves are preserved by reference; subtree-0 coinbase
  placeholder handling and `correctedSubtree0Root` are unchanged.
  - New internal helper `computeAndPadCompound` does the top-down climb in a single pass using
  **per-level offset maps** for O(1) sibling/parent-existence lookups (replacing repeated O(N)
  `findLeafByOffset` linear scans) and folds the Bitcoin-canonical "duplicate last on odd" padding
  into the same loop for the block-level layers.
  - `services/bump_builder` is untouched — same caller, no API change.

  ## Why It Was Necessary

  The old algorithm's cost was **O(N²·log N) time and O(N²) intermediate `PathElement` allocations**
  in the subtree count `N`. The hottest contributors:

  - `assembleFullPath` called N times — each call built the full block-level tree skeleton
  (subtree-root layer + log₂(N) padding levels) just to be merged away.
  - `computeMissingHashes` ran inside each of those calls; `findLeafByOffset` is a linear scan, so
  each level is O(M²) — paid N times.

  This was invisible at the few-thousand-subtree shapes BSV blocks used to have, and unavoidable at
  Teranode's tens-of-thousands-of-subtree shapes. Today's mainnet routinely produces blocks in the
  latter range; without this fix, arcade's bump-builder OOMs (and DLQs the block-processed message) on
   the first such block that contains arcade-tracked txs after a deploy.

  ## Testing Performed

  - **All 9 existing `BuildCompoundBUMP` + `ValidateCompoundRoot` tests pass unchanged** — these are
  the byte/semantic equivalence proof. Same merkle root, same compound shape, same minimal-path
  extraction for tracked leaves. (`TestBuildCompoundBUMP_AllTxsExtractable`, `_CoinbaseReplacement`,
  `_CoinbaseReplacement_SingleSubtree`, `_NonPow2SubtreeCounts`, `_39Subtrees_StructuralShape`,
  `_NonPow2_WithCoinbase`, `_CoinbaseReplacement_OrderIndependence`, `_NoSubtree0Stump_NonPow2`, plus
  `TestValidateCompoundRoot`.)
  - `ExtractMinimalPath` and `AssembleBUMP` tests pass unchanged.
  - Full `./bump/...` and `./services/bump_builder/...` suites green.
  - New `TestBuildCompoundBUMP_LargeBlock_DoesNotOOM` — 4,096 subtrees × 16 leaves (65,536 block
  leaves) builds in **~0.13 s**, verifies against the independently-computed canonical root, and
  spot-checks `ExtractMinimalPath` round-trips for the first and last leaves of several subtrees. The
  old algorithm could not have completed in any reasonable time on the same input — this test will
  hard-regress (timeout / order-of-magnitude slowdown) if the algorithm ever slips back to per-STUMP
  merging.

  ## Impact / Risk

  - **No API change.** Same exported signature; bump-builder, store, and downstream merkle-path
  serving are untouched.
  - **Output is structurally equivalent**, validated by every existing test. The only path that
  wouldn't survive a regression here is the production bump-builder, and that's covered by the
  bump_builder integration suite.
  - The single-subtree edge case still delegates to `assembleFullPath` (unchanged code path) — the new
   direct-build only kicks in for multi-subtree blocks.
  - Bisecting back is a one-function revert if something is missed; no schema, no Kafka, no config
  touched.
  - Related (out of scope for this PR): the `feat/backfill-recovery-tool` branch on this fork
  addresses a different failure mode — recovering BUMPs for blocks where merkle-service never
  delivered any STUMPs at all (every known datahub had pruned the subtree data). That tool ships independently and is not on this PR's
  critical path.

### DIFF
--- a/bump/bump.go
+++ b/bump/bump.go
@@ -442,9 +442,30 @@ func correctedSubtree0Root(stumps []*models.Stump, coinbaseTxID *chainhash.Hash,
 	return subtree0RootFromCoinbaseBUMP(coinbaseBUMP, len(mp.Path))
 }
 
-// BuildCompoundBUMP merges multiple per-subtree BUMPs into a single compound MerklePath
-// containing all tracked transactions at level 0. The tracker is used to discover
-// which level-0 hashes are tracked transactions.
+// BuildCompoundBUMP merges multiple per-subtree BUMPs into a single compound
+// MerklePath containing every block leaf the input STUMPs cover (with each
+// STUMP's tracked-txid markers preserved). The block-level tree is filled in
+// from the STUMP leaves and the supplied subtreeHashes.
+//
+// This is the live arcade hot path. The previous implementation called
+// assembleFullPath once per STUMP — each call building the entire block-level
+// tree (subtree-root layer + log₂(N) climb layers) and then deduping them all
+// through a (level, offset) map. For a tens-of-thousands-of-subtree block
+// that path allocated N copies of the block-level tree (tens of GB of
+// intermediate PathElement pointers) and ran computeMissingHashes once per
+// STUMP at quadratic cost, so the bump-builder would OOM rather than complete
+// (verified 2026-05-22 against block 950028: 24,896 STUMPs drove arcade to
+// 13 GB / 100% CPU with no completion after 8+ min).
+//
+// The current implementation builds the compound directly: each STUMP's
+// elements are added at their global block offsets (one pass per STUMP, no
+// per-STUMP climb), the subtree-root layer is seeded once with subtreeHashes
+// for slots that do not have a STUMP, and the rest of the tree is computed
+// in a single top-down pass with per-level offset maps so each level is O(N)
+// instead of the O(N²) findLeafByOffset behaviour used by computeMissingHashes.
+// Output is structurally equivalent to the old algorithm — same merkle root,
+// same extracted minimal paths for tracked txs — and the nine existing
+// BuildCompoundBUMP tests pass unchanged.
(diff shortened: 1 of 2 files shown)


galt-tr:
Thanks @mboyd1 ; pulling this in to test

mrz1836 (approved):
LGTM

galt-tr (approved):

PR #168, opened 2026-05-25
fix(bump): climb single-subtree compound BUMP to block root (#167)

### LINKED ISSUE #167: [Bug] Compound BUMP for single-subtree blocks computes wrong merkle root → txs stuck at SEEN_MULTIPLE_NODES, never reach MINED
### Describe the bug

For at least some **single-subtree blocks**, `bump.BuildCompoundBUMP` produces a compound BUMP whose computed merkle root does **not** match the canonical block-header merkle root. The bump-builder's post-build guard (`ValidateCompoundRoot`) correctly rejects it and refuses to persist, so:

- no compound BUMP is ever stored for the block,
- the block's tracked transactions are never moved to `MINED`,
- they remain stuck at `SEEN_MULTIPLE_NODES` indefinitely, even though the block is mined and confirmed on-chain.

Because the failure is deterministic, **`/reprocess` cannot recover it** — every reprocess re-runs the same broken computation. The transactions never recover without a code fix.

Observed in production (HandCash arcade deployment, BSV mainnet):

- Block height **950675**, hash `00000000000000000eb6ea115a5cb140ffeae7b673b4c5b6f3fac62e0adae3c5`
- Block shape (from the failure-dump logs): **`subtree_count: 1`**, **`stump_count: 1`** (subtree-0 STUMP present, ~4851 bytes), **`txid_count: 40`** tracked txids.
- Canonical block-header merkle root: `32d293e3c4947fee702128ff1552c85d2959880e03f7f225600ed03ae240a202`
- Computed compound root (**v0.6.7**): `e0360bc5c3163991db5c116414d51d51436297a0f08e9dfab88fb235767cb495` ❌

The two versions compute **different** wrong roots, neither matching the header. This is consistent with the divergence in `bump/bump.go` between the two builds: **v0.6.7 has a dedicated single-subtree branch inside `BuildCompoundBUMP` (bump.go:492) that the older branch lacks** — but the extra logic still does not reconcile to the header root.

Not a one-off: in the same window the `de50a2c` build also failed a **second, distinct** block — computed `975619c0…` vs header `1fcceb08…` — so this is **systemic for the single-subtree block shape**, not a single bad block.

Example stuck transaction: `cd0f489ee420379537fc6dcdf54434ba86e0267ca039c551b0d37d93d7034bbf` → still `SEEN_MULTIPLE_NODES`, no `blockHash`/`merklePath`, despite being mined in block 950675.

### Steps to reproduce

1. Submit/track transactions that land in a **single-subtree** mainnet block (`subtree_count == 1`) with a full subtree-0 STUMP and a non-trivial tracked-tx count (observed: 40).
2. Let the block be processed (or trigger `POST /api/v1/blocks/<blockHash>/reprocess`).
3. Observe in the bump-builder logs:
   - `building compound BUMP`
   - `compound BUMP validation failed — refusing to persist`
   - `compound BUMP root mismatch for block <hash>: computed root <X> != block header merkle root <Y>`
4. `GET /tx/<txid>` shows the tx remains `SEEN_MULTIPLE_NODES`; `GET /api/v1/blocks/processing-status/<hash>` shows `hasBlockProcessed: true, hasCompoundBUMP: false`.

Concrete reproduction values: block `00000000000000000eb6ea115a5cb140ffeae7b673b4c5b6f3fac62e0adae3c5` (height 950675), header merkle root `32d293e3…40a202`.

**Likely test gap:** the existing single-subtree coverage (`TestBuildCompoundBUMP_CoinbaseReplacement_SingleSubtree`, `TestAssembleBUMP_SingleSubtree_*`) only exercises tiny **power-of-two** synthetic subtrees (4/8/16 leaves) with a hand-built coinbase BUMP, and validates by re-parsing the compound — not against a datahub-supplied header root for a realistic single-subtree mainnet block. A regression test built from the block above (or any real single-subtree block) would reproduce it.

### Expected behavior

`BuildCompoundBUMP` for a single-subtree block produces a compound whose `ComputeRoot` equals the block-header merkle root (`32d293e3…40a202`), `ValidateCompoundRoot` passes, the compound BUMP is persisted, and the block's tracked transactions transition to `MINED` with a valid `merklePath`.

### Environment details

- Library/service: `github.com/bsv-blockchain/arcade`, `bump` package + `services/bump_builder`.
- Affected version: **v0.6.7** (release, computes `e0360bc5…`)
- Network: BSV **mainnet**.
- Datahub sources: multiple Teranode datahubs (all healthy at the time).
- Deployment: EKS, `--mode all`, 2 replicas

### Additional context

Log evidence (CloudWatch, `compound BUMP root mismatch`):

```
15x  image=arcade:v0.6.7         computed=e0360bc5…  header=32d293e3…   (block 950675)
```

Block-processing row at time of failure:

```json
{
  "blockHash": "00000000000000000eb6ea115a5cb140ffeae7b673b4c5b6f3fac62e0adae3c5",
  "blockHeight": 950675,
  "headerSeenAt": "2026-05-25T17:42:08.034Z",
  "processedAt":  "2026-05-25T17:42:08.417Z",
  "hasBlockProcessed": true,
  "hasCompoundBUMP": false,
  "status": "active"
}
```

Impact: affected transactions are mined and safe on-chain, but arcade can never emit a BUMP/merkle proof for them or report `MINED`. The watchdog does **not** recover this case (it only re-fires `/reprocess` for blocks with `header_seen_at` but no `processed_at`; here `processed_at` is set, so the block is never flagged as stale).

rafa-js:
## What Changed
- `BuildCompoundBUMP`'s single-subtree path now folds the subtree root up to the true block-header merkle root using the coinbase BUMP's left-spine siblings (new `extendSubtreeToBlockRoot` helper in `bump/bump.go`).
- Added a regression test (`bump/issue167_test.go`) plus the real captured inputs as fixtures (`bump/issue167_fixtures_test.go`) from mainnet block 950675.

## Why It Was Necessary
- A single-subtree block whose coinbase sits outside the subtree has a block tree taller than the subtree STUMP (for block 950675: block height 16 vs STUMP internal height 15, 39,657 txs). The old single-subtree shortcut treated the STUMP as the full block path and returned the **subtree** root instead of the **block** root.
- `ValidateCompoundRoot` therefore rejected the compound (computed `e0360bc5…` ≠ header `32d293e3…`), no BUMP was persisted, and the block's tracked transactions were stuck at `SEEN_MULTIPLE_NODES` and never reached `MINED`. The failure is deterministic, so `/reprocess` could not recover it.
- Fixes #167.

## Testing Performed
- New test reproduces the production failure (computed `e0360bc5…` ≠ header `32d293e3…`) **before** the fix and passes **after**; it also verifies every tracked tx's minimal path computes the canonical block-header merkle root `32d293e3…`.
- `go test ./bump/` and `go test ./services/bump_builder/`: pass. `go build ./...`: clean.
- `golangci-lint` v2.12.2 (CI-pinned) on `./bump/`: 0 issues. `gofmt`: clean.

## Impact / Risk
- Low. The new fold is a no-op when the coinbase BUMP is not taller than the STUMP, so power-of-two single-subtree blocks and all multi-subtree blocks are unaffected (the full `bump` suite still passes). Only previously-failing single-subtree blocks change behaviour — from "rejected, no BUMP" to a valid, header-reconciling compound BUMP.

## Notifications
- @mrz1836


galt-tr:
Thank you, good catch. I am surprised we didn't catch this before; I see many single subtree blocks on TTN where this isn't the case. Wondering if theres something unique about how teranode stores blocks from a legacy sv node?

regardless, I can confirm this is a real bug.

galt-tr (approved):

PR #170, opened 2026-05-26
Make webhook delivery idempotent

galt-tr:
Fixes #166 
This pull request implements exactly-once webhook delivery across horizontally-scaled replicas by introducing a compare-and-swap (CAS) mechanism for updating delivery status. The main change is the addition of the `UpdateDeliveryStatusCAS` method to the store interface and its implementation across all supported backends (Postgres, Aerospike, Pebble), ensuring that only one replica processes each webhook event. The PR also adds metrics, regression tests, and updates existing tests to reflect the new delivery semantics.

**Exactly-once delivery and CAS coordination:**

* Added `UpdateDeliveryStatusCAS` to the `store.Store` interface and implemented it for Postgres, Aerospike, and Pebble. This method atomically advances `LastDeliveredStatus` only if its current value matches the expected state, preventing duplicate webhook POSTs from concurrent replicas. [[1]](diffhunk://#diff-9a62b4a044185786fb1bf1c5a2ea3d0d94c2e168d1ed580db8994c06a4f0f41aR214-R226) [[2]](diffhunk://#diff-244a4cf665c25ac2ad8350a33a471e9f23a4d6ea0191d32c67b1c1c75ff9092fR1134-R1153) [[3]](diffhunk://#diff-89f39afc0684c78f0192c7967cdaf1df5c2a38a5983f15e3b7130a7f0e4fa331R1583-R1625) [[4]](diffhunk://#diff-b054e9edce44635c35073f374e2a171b37a6b47a83fb998a308ef3c17ef33147R54-R58) [[5]](diffhunk://#diff-b054e9edce44635c35073f374e2a171b37a6b47a83fb998a308ef3c17ef33147R1560-R1604)
* Updated the webhook delivery logic in `service.go` to use `UpdateDeliveryStatusCAS` for cross-replica coordination, ensuring only one POST per event and tracking CAS-lost attempts via a new metric. [[1]](diffhunk://#diff-992015a8c9b6e512379c1085c2a3253e81a6a5943af179164d9fbbe2707f01feL262-R292) [[2]](diffhunk://#diff-992015a8c9b6e512379c1085c2a3253e81a6a5943af179164d9fbbe2707f01feL302-R322) [[3]](diffhunk://#diff-d715f391f2ffa038c3167bfea1adf741c1f5e39f08349fbbacab780a594146a4R378-R390)

**Metrics and observability:**

* Introduced the `WebhookCASLostTotal` Prometheus metric to count CAS-lost attempts, providing visibility into cross-replica coordination and deployment health.

**Testing and regression coverage:**

* Added and updated tests to verify exactly-once semantics, including a regression test for concurrent delivery attempts and backend-specific tests for CAS correctness. [[1]](diffhunk://#diff-79b7cf0ab7d67e7059a717d4a02c20c791f43c1206274cfd98c976a4f5ea5cb0R565-R628) [[2]](diffhunk://#diff-f4d65b20a524391775206eb9c23524c19604a874517d4c5922efb32b71ef89d7R415-R491) [[3]](diffhunk://#diff-79b7cf0ab7d67e7059a717d4a02c20c791f43c1206274cfd98c976a4f5ea5cb0R69-R96) [[4]](diffhunk://#diff-fdc60e100b1149e61c9d0c41538e83d0df51c2431aad320e7d02ee0cd41c9ca6R178-R181)
* Updated existing tests to account for the new CAS claim and retry bookkeeping, ensuring correct delivery record counts and state transitions. [[1]](diffhunk://#diff-79b7cf0ab7d67e7059a717d4a02c20c791f43c1206274cfd98c976a4f5ea5cb0L315-R350) [[2]](diffhunk://#diff-79b7cf0ab7d67e7059a717d4a02c20c791f43c1206274cfd98c976a4f5ea5cb0L406-R444)

**Backend-specific improvements:**

* Ensured that all store backends correctly read and write `LastDeliveredStatus` and related fields, fixing potential deduplication and state tracking issues.

These changes collectively address issue #166, ensuring that webhook POSTs are delivered exactly once per event, even in the presence of multiple concurrently processing replicas.


rafa-js (commented):
Nice work! the CAS approach is the right call here. Routing concurrent attempts through a conditional `last_delivered_status` advance is structurally exactly-once at any replica count, and it also kills the rolling-deploy `maxSurge` duplicate without special-casing rollouts. Leader election (the lease/single-active route from the issue) would have been the heavier and, given the point below, less correct option, since the standby would drop events during failover with nothing to re-scan them. Glad this went the CAS way.

A few things I like beyond the core fix:

- The `recordToSubmission` change in Aerospike looks like it fixes a real latent bug. `last_status` wasn't being read back, so `LastDeliveredStatus` was always the zero value and dedup never actually worked on that backend. Good catch.
- `WebhookCASLostTotal` is a useful signal. A flat zero on a multi-replica deployment is a nice "events aren't flowing through CAS" tripwire.
- Postgres first-delivery handling (`expected='' ` matching `NULL` or `''`) and the concurrency regression test are exactly what I'd want to see.

## One blocking concern: the failure path is now a regression, not just an unfixed gap

The PR advances `LastDeliveredStatus` at **claim time, before the POST**, and `recordFailure` deliberately writes the post-CAS value (per the comment warning against rollback). That's defensible on concurrency grounds, but combined with the fact that **nothing reads `NextRetryAt` back** (there's no webhook retry sweep; `GetReadyRetries`/`PendingRetry` serve the propagation reaper, a different concern), it makes a failed delivery strictly worse than before this PR:

- **Before:** a failed POST left `LastDeliveredStatus` at the old value. If the same status event re-arrived, `shouldDeliver` passed (`old != new`) and re-attempted. Event-driven retry actually worked.
- **After:** the CAS advances `LastDeliveredStatus` to the new value before the POST. On failure it stays advanced, so a re-arriving event hits `shouldDeliver`'s `LastDeliveredStatus == status.Status → return false`. The one retry mechanism that worked is now disabled, and the `RetryCount`/`NextRetryAt` that `recordFailure` writes are never consumed by anyone.

Net effect: a callback that fails its single MINED POST is now silently dropped forever. The `deliver` doc comment already acknowledges this ("the retry path needs a separate event or sweep to re-fire, see the follow-up issue"). My worry is that the follow-up is what makes this whole change durable, so deferring it means shipping a known delivery-loss path.

Two ways to resolve, either is fine:

1. **(Preferred) Land the sweep in this PR.** A `ListSubmissionsReadyForRetry(now, limit)` store method plus a lease-gated ticker that re-fires due deliveries. This also retroactively fixes the pre-existing single-replica "failed deliveries are never retried" behavior, so it's worth doing regardless.
2. **Treat the deferred-sweep issue as a merge blocker / fast-follow**, explicitly, rather than a backlog nicety, and link it here so the gap is tracked against this change.

## Minor / confirm before merge

- Postgres migration: is the `last_delivered_status` column actually present in the deployed schema? If the migration isn't in place the CAS silently matches nothing (or errors), and `WebhookCASLostTotal` staying flat-zero would be the only symptom. Worth confirming the migration ships with this.
- The Aerospike impl treats a generation mismatch and a transient infra error identically (both return `claimed=false, nil`). That's safe for correctness, but it means real infra errors are invisible at this layer and only recoverable if another event arrives, which ties back to the retry-sweep gap above. A debug log on the non-nil-but-swallowed case might help operability.

rafa-js (approved):
Re-reviewed after the `Address feedback` push (`945817e`). The blocking concern is fully resolved, approving.

## Blocking concern resolved (option 1: land the sweep)

The failure path is no longer a regression. The new lease-gated `services/webhook/reaper.go` (mirroring `services/propagation/reaper.go`: run-once-on-entry, batch cap, best-effort lease release on shutdown) plus the new `ListSubmissionsReadyForRetry` store method across all three backends now consume the `retry_count`/`next_retry_at` bookkeeping that `recordFailure` writes. A transient 5xx is retried instead of silently dropped.

The CAS interplay is correct: the reaper synthesizes `status = LastDeliveredStatus`, so the re-fire CAS is a no-op on status while clearing retry state, and the captured `sub.RetryCount` preserves the backoff counter so escalation continues. Terminal exhaustion writes `next_retry_at = NULL`, which the list query excludes, so dead submissions stop being swept.

`recordFailure` now writes `status.Status` (the post-CAS value) rather than the pre-CAS in-memory value, closing the rollback hazard called out earlier.

## Minor points, both addressed

- Postgres migration: `schema.sql` adds idempotent `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` for the three webhook columns, plus the `idx_sub_retry_ready` partial index keyed to the reaper scan predicate. The deployed-schema gap is covered.
- Aerospike gen-mismatch vs real infra error are now distinguishable via the dedicated `WebhookCASErrorTotal` metric, which is stronger than the debug log I suggested.

## Verified locally

- `go build ./...` clean; `go test ./services/webhook/...` and `./store/pebble/...` pass; Postgres integration tests (`-tags=postgres`) pass.
- `TestDeliver_ExactlyOnceAcrossConcurrentReplicas` races two Service instances through a shared mutex-guarded store and asserts exactly one POST, a genuine concurrency regression test.

Nice work landing the sweep here rather than deferring it, it also retroactively fixes the pre-existing single-replica "failed deliveries never retried" behavior.

mrz1836 (approved):
TLDR; LGTM

PR #171, opened 2026-05-26
fix: broadcast aggregation, validation, and REJECTED recovery

shruggr:
## Summary

Four related correctness fixes to the propagation and validation paths. Each commit is reviewable independently.

- **models**: REJECTED can transition forward to ACCEPTED_BY_NETWORK / SEEN_ON_NETWORK / SEEN_MULTIPLE_NODES so a late peer acceptance can correct an earlier rejection. Backward regressions and the DOUBLE_SPEND_ATTEMPTED lattice are unchanged.
- **propagation: stop cancelling sibling broadcasts on first success**. The result loop already runs to completion regardless of cancellation, so the early-cancel only aborted in-flight HTTP calls to other Teranodes, leaving them without the tx. That gap caused downstream "missing input" rejections on dependent broadcasts.
- **propagation: aggregate per-peer responses**. Replaces the "first parseable failure map wins" logic with per-tx accumulation. Acceptance from any peer wins for that tx (sticky). A tx is only REJECTED when every responding peer explicitly listed it as failed.
- **validator: enable script and fee validation at intake**. Flips the two handler call sites from `skipFees=true, skipScripts=true` to `skipFees=false`. spv.Verify runs script execution against per-input EF/BEEF source data, catching malformed unlocking scripts, bad signatures, and underpaid txs at intake with HTTP 400. Merkle proofs on BEEF parents are still trusted (no chaintracker on the intake path, no external fallback). The unused skipScripts parameter and ChainTracker plumbing in the validator are dropped.

## Test plan

- [x] `go test ./...` — full suite passes
- [x] New propagation tests: `TestBroadcast_PerPeerAggregation_AcceptanceWins`, `TestBroadcast_PerPeerAggregation_UnanimousRejection`, `TestBroadcastSingle_WaitsForSlowPeer`
- [x] New models test: `TestStatus_CanTransitionFrom_RejectedRecovery` plus updated `TerminalStaysTerminal`
- [x] New store/pebble test: `TestUpdateStatus_RejectedRecovery` plus updated `TerminalNotOverwritten`
- [ ] Deploy to a non-prod environment and confirm Teranode broadcast hit counts increase per chunk (every healthy endpoint now receives every tx)
- [ ] Confirm intake validation rejects raw tx submissions (no source data) with a 400 carrying `missing source transaction` in the reason


models/transaction.go
```diff
@@ -150,24 +150,24 @@ func (s Status) DisallowedPreviousStatuses() []Status {
...
 			StatusSeenOnNetwork, StatusSeenMultipleNodes,
-			StatusRejected, StatusDoubleSpendAttempted,
+			StatusDoubleSpendAttempted,
 			StatusMined, StatusImmutable,
 		}
 	case StatusSeenOnNetwork:
-		// SEEN_ON_NETWORK is a forward step from earlier states only — once a
-		// tx has progressed to SEEN_MULTIPLE_NODES, a terminal state, or is in
-		// the retry side-branch it must not regress to SEEN_ON_NETWORK.
+		// SEEN_ON_NETWORK is a forward step that can follow earlier states or
```
rafa-js, line 157: I'm not sure that interpretation is as simple as that. Can you elaborate more about this idea?

shruggr: Previously, if a transaction was marked as rejected, and was later seen in a subtree, it was not allowed to transition to SEEN_ON_NETWORK, even though it was legitimately seen on the network. 

REJECTED was treated as a terminal state, even though it might have only been rejected for the user having broadcast transactions in the wrong order, where a child is broadcasted before a parent. As it was, resubmitting a rejected transaction would NOT rebroadcast and even if it did get relayed to a node, it's state would never update.

galt-tr: Agree; I've encountered the same behavior.

tests/e2e/fixtures/blocks/000000000000000001bc8a601dd5f0659d36a9b077808850375dfa2d9f009396/meta.json
```diff
@@ -1,28 +1,28 @@
...
-    "txCount": 1911
+  "blockHash": "000000000000000001bc8a601dd5f0659d36a9b077808850375dfa2d9f009396",
+  "height": 948351,
+  "merkleRoot": "215b459eb39fb46220c591d89dffa89def1cf03f327765e6009b949aed741414",
+  "txCount": 1911,
+  "subtreeCount": 1,
+  "subtrees": [
+    "dc86d92dd782fa56d4c01aa1e20d1e474eed53ffa9c05a4d4b67cdda49f6c7bd"
+  ],
+  "coinbaseTxID": "30e099b7318973c83020132c6b1b77f0a6bade9a4d28d36c004076d8a674d7c6",
```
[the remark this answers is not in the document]
shruggr: This is not a secret

models/transaction.go
rafa-js: Agree too; I have encountered the same.

I'm trying to think if there is a negative case: REJECTED -> SEEN_ON_NETWORK introduces a false positive.

mrz1836 (approved):
TLDR; LGTM

models/transaction.go
galt-tr: Shouldn't be because merkle-service only sends this callback when a tx is in a subtree it encounters on the network

galt-tr (approved):

PR #172, opened 2026-05-26
fix(bump): skip datahub peers whose coinbase BUMP does not reconcile (#167)

rafa-js:
## What Changed
- `FetchBlockDataForBUMPWithOptions` now rejects a datahub endpoint whose coinbase BUMP does not compute the header merkle root it reports, via a new `coinbaseBUMPReconciles` check in the fetch loop (`bump/datahub.go`).
- Added tests using the real consistent/inconsistent coinbase BUMPs captured for mainnet block 950675, plus an integration test proving the loop skips the inconsistent peer and falls through to a consistent one.

## Why It Was Necessary
- A stale/buggy datahub peer can report the correct block-header merkle root while serving a coinbase BUMP (and subtree decomposition) that computes a different root. Such a peer passes the fetch-time validator (`chainHeaderRootValidator` only checks the *reported* root).
- If that peer sorts ahead of a consistent one, the fetch loop selects it on every attempt and every Kafka retry, so the build deterministically fails the post-build `ValidateCompoundRoot` and the block can never produce a valid BUMP, even when a consistent peer is healthy. This left block 950675's tracked transactions stuck at `SEEN_MULTIPLE_NODES`.
- A coinbase BUMP is the coinbase tx's path to the block root, so it MUST compute the header merkle root; verifying that lets the loop skip inconsistent peers. Complements #168 (single-subtree compound correctness): together they let 950675 build from a consistent peer and reach `MINED`.

## Testing Performed
- `TestCoinbaseBUMPReconciles_RealData`: the consistent coinbase BUMP reconciles, the inconsistent one (computes `0a4d286f…` vs real `32d293e3…`) is rejected; nil/empty inputs are no-ops.
- `TestFetchBlockData_SkipsInconsistentCoinbaseEndpoint`: with the bad peer first, the loop returns the good peer's data; with only the bad peer, it surfaces a `coinbase BUMP mismatch` error.
- `go test ./bump/` and `./services/bump_builder/`: pass. `go build ./...`: clean. `golangci-lint` v2.12.2 on `./bump/`: 0 issues. `gofmt`: clean.

## Impact / Risk
- Low. The check is a no-op when a peer omits the coinbase BUMP, so endpoints that legitimately return no coinbase BUMP are unaffected. It only adds a rejection reason in the existing fetch fall-through loop (same mechanism as `SubtreeCountValidator`), and adds one cheap `ComputeRoot` per fetched block. Endpoints serving consistent data are selected exactly as before.

## Notifications
- @mrz1836


galt-tr (approved):
I am surprised this is happening on the network with the current datahub URLs; but in general this is a good check to have because a malicious peer could be active as well.

This might require some digging in teranode code though to figure out how this could be happening

rafa-js:
Sharing the evidence that led to this PR.

# RCA: datahub peers serve inconsistent subtree data, blocking BUMP construction

- Service: arcade (bump-builder).
- Network: BSV mainnet.
- Production instance: arcade.handcash.io.
- Reference block: 950675.

## Problem statement

Different datahub peers return materially different block representations for the
same block. Some peers serve a subtree decomposition and coinbase BUMP that do
not reconcile to the block-header merkle root. Because arcade's fetch-time
validator only checks the header root a peer reports (not whether the peer's
subtree/coinbase data actually computes it), an inconsistent peer is accepted,
and the post-build `ValidateCompoundRoot` then rejects the result. When an
inconsistent peer sorts ahead of a consistent one, arcade selects it on every
fetch and every retry, so the block can never build a valid BUMP even though a
consistent peer is healthy.

PR #172 closes this by verifying, in the fetch loop, that a peer's coinbase BUMP
computes the header merkle root it reports, and skipping peers that fail.

## Primary evidence: per-peer probe of block 950675

Block `00000000000000000eb6ea115a5cb140ffeae7b673b4c5b6f3fac62e0adae3c5`
(height 950675, 39,657 txs). Canonical block-header merkle root, confirmed
independently via WhatsOnChain:

```
32d293e3c4947fee702128ff1552c85d2959880e03f7f225600ed03ae240a202
```

Each configured datahub was probed by replaying arcade's own `fetchBlockBinary`
and computing the root of the coinbase BUMP it returned (the coinbase tx is leaf
0 of the block tree, so its BUMP MUST compute the header root):

| Datahub peer | reported header root | subtree_count | coinbase BUMP root | reconciles |
|---|---|---|---|---|
| `teranode-eks-mainnet-eu-1.bsvb.tech` | `32d293e3...` | **1** | `32d293e3...` | **yes** |
| `mainnet.gorillanode.io` | `32d293e3...` | **2** | `0a4d286f...` | **no** |
| `bsva-ovh-teranode-eu-1.bsvb.tech:8000` | `32d293e3...` | **2** | `0a4d286f...` | **no** |
| `mainnet2.gorillanode.io` | n/a | n/a | n/a | HTTP 502 |
| `bsva-ovh-teranode-eu-3.bsvb.tech:8000` | n/a | n/a | n/a | not reachable from probe host |
| `teranode-mainnet.orches-trator.com` | n/a | n/a | n/a | not reachable from probe host |
| `bsva-ovh-teranode-eu-2.bsvb.tech` | n/a | n/a | n/a | not reachable from probe host |

The unreachable peers are reachable in-cluster; they were firewalled from the
probe host. The reachable set is sufficient to prove the inconsistency.

### Inconsistent subtree decomposition (the core finding)

The same block is decomposed differently by different peers:

- **Consistent peer** (`teranode-eks-mainnet-eu-1`): **1 subtree**
  - `subtree[0] = 468f4ed8ff40546e2c6b7f6e7843bed5d7d088cc75c5c05bb42e3b0731e86edc`
  - Its coinbase BUMP computes `32d293e3...` (matches the canonical root).

- **Inconsistent peers** (`mainnet.gorillanode.io`, `bsva-ovh-eu-1`): **2 subtrees**, identical to each other:
  - `subtree[0] = e3287a40f82c8ff7395668b1b3f247dcdcc4ae18dd6b913c090f09e3d7a10c91`
  - `subtree[1] = 05880cf5a9009d3162399936422c647ea2106b88561034159444c1dda9d90b9e`
  - Their coinbase BUMP computes `0a4d286f...`, which does NOT match the
    canonical root, even though they report `32d293e3...` as the header root.

Two facts make the inconsistent peers' data provably wrong:

1. **Self-inconsistency.** A coinbase BUMP is the coinbase tx's path to the block
   root, so it must compute the header merkle root the same peer reports. These
   peers report `32d293e3...` but their coinbase BUMP computes `0a4d286f...`.
2. **Disagreement with the canonical chain.** `0a4d286f...` is not block 950675's
   merkle root (WhatsOnChain confirms `32d293e3...`).

Note: `subtree[0] = e3287a40...` on the inconsistent peers is the subtree-0 root
computed against the coinbase placeholder (the uncorrected value), which further
indicates these peers are serving a stale or pre-coinbase-correction
representation of the block.

## Why this blocked recovery (deterministic peer selection)

1. All peers, consistent and not, **report the correct header root**
   (`32d293e3...`). arcade's fetch-time validator (`chainHeaderRootValidator`)
   compares only this reported root against chaintracks, so inconsistent peers
   pass it.
2. `FetchBlockDataForBUMPWithOptions` returns the first healthy peer in list
   order. When an inconsistent peer sorts first, it is selected.
3. The build then fails the post-build `ValidateCompoundRoot` (computed root
   `0a4d286f...` != header `32d293e3...`), so arcade refuses to persist and the
   Kafka message errors and retries.
4. The retry re-fetches in the same order and selects the same inconsistent peer
   again. A deterministic loop on a bad peer, with no progress, even though
   `teranode-eks-mainnet-eu-1` was healthy and consistent the whole time.

### Production log signature (CloudWatch, bump-builder)

```
msg: "building compound BUMP"
msg: "compound BUMP validation failed - refusing to persist"
err: "compound BUMP root mismatch for block 0000...adae3c5: computed root 0a4d286f... != block header merkle root 32d293e3..."
msg: "message processing failed, retrying"
```

Repeated `POST /reprocess` only advanced `processedAt`; `hasCompoundBUMP`
remained `false`. This confirmed the failure was deterministic, not transient.

## Validation gap

Nothing verified that the fetched coinbase BUMP (and therefore the subtree
decomposition that drives the build) actually computes the reported header root.
The fetch-time validator trusted the reported header root; the only check that
catches inconsistent data was the post-build `ValidateCompoundRoot`, and a
post-build failure did not cause the fetch to try a different peer.

## Operational follow-up (datahub fleet)

The inconsistent data is a datahub fleet problem independent of arcade:

- `mainnet.gorillanode.io` and `bsva-ovh-teranode-eu-1` served a subtree
  decomposition and coinbase BUMP for block 950675 that do not reconcile to the
  block they claim. They should be investigated (stale cache, pruned block,
  pre-coinbase-correction representation) and either fixed or de-prioritized.
- `mainnet2.gorillanode.io` returned HTTP 502 for this block.
- PR #172 makes arcade robust to these peers, but the peers serving wrong data
  for confirmed blocks is worth raising with the datahub operators.

## How to reproduce the evidence

For any reachable datahub `BASE`:

```
GET {BASE}/block/00000000000000000eb6ea115a5cb140ffeae7b673b4c5b6f3fac62e0adae3c5
```

Parse the binary block format (header | txCount | sizeBytes | subtreeCount |
subtreeHashes | coinbaseTx | blockHeight | coinbaseBUMPLen | coinbaseBUMP),
compute the coinbase BUMP root from the coinbase txid, and compare to the
canonical root `32d293e3...`. Consistent peers match; inconsistent peers do not.

PR #184, opened 2026-05-30
[bump] Derive corrected subtree-0 root from coinbase BUMP, not over-climbed STUMP

rafa-js:
## What Changed

- Delete `computeCorrectedSubtreeRoot`, whose `stumpPath.ComputeRoot(coinbaseTxID)` climbed every level of the subtree-0 STUMP with no height bound.
- Rewrite `correctedSubtree0Root(coinbaseBUMP []byte, numSubtrees int)` to derive the corrected subtree-0 root solely from the coinbase BUMP: `internalHeight = len(cbPath.Path) − ceil(log2(numSubtrees))`, then fold the coinbase exactly `internalHeight` levels via the existing bounded primitive `subtree0RootFromCoinbaseBUMP`.
- In `BuildCompoundBUMP`, derive the placement `internalHeight` from the coinbase BUMP when present (same formula), falling back to the first STUMP's height only when no coinbase BUMP is available.
- Relax the placement height guard from `!=` to `<` so an over-tall STUMP is tolerated (only its first `internalHeight` levels are read); a STUMP shorter than `internalHeight` is still rejected.
- Add reproduction tests at the block-951360 shape and re-point the surviving unit test at `subtree0RootFromCoinbaseBUMP`.

## Why It Was Necessary

merkle-service sometimes delivers subtree 0's STUMP at full block height instead of the subtree-internal height. The old unbounded climb then overshoots the subtree-0 root and lands on the block merkle root, poisoning `subtreeHashes[0]`; the compound's top tree is built over `[block_root, s1, s2, …]` and `ValidateCompoundRoot` rejects it. The transaction is never marked `MINED`, so it remains stuck and gets rebroadcast. Observed on mainnet block 951360 (tx `85598b7b2d994153487198f9b69b482471d496c50e694f17bb92dd611edadc97`, `subtreeCount = 3`): block root `5ed5a7a9…ecbc`, wrong computed root `49c8fd3a…4f23`, STUMP `treeHeight = 17` vs. true `internalHeight = 15`. The fix derives `internalHeight` from the coinbase BUMP (authoritatively full block height), making the correction immune to over-tall STUMPs.

Closes #183

## Testing Performed

- `go test ./bump/...` — 89 tests pass, including two new reproductions: `TestBuildCompoundBUMP_FullBlockSubtree0STUMP_Block951360` (over-tall subtree-0 STUMP + normal siblings; validates against the block root) and `TestCorrectedSubtree0Root_Block951360_FromCoinbaseBUMP` (corrected root equals the real subtree root, not the block root).
- TDD: both reproductions fail on the unmodified tree (placement height-mismatch / over-climb poison) and pass after the fix.
- `go vet ./bump/...` and `gofmt -l bump/` are clean.

## Impact / Risk

- **Breaking change:** None — public signatures of `BuildCompoundBUMP` / `AssembleBUMP` are unchanged.
- **Behavioral risk:** Low. For well-formed inputs the derived `internalHeight` equals the previous STUMP-derived value, so existing behavior is preserved; the change only affects over-tall STUMPs that previously failed.
- **Caveat:** The correction relies on a coinbase BUMP being present for multi-subtree blocks. With no coinbase BUMP, both the correction and the placement-height derivation fall back to the STUMP height (today's behavior), so a `coinbaseBUMP == nil` path with an over-tall STUMP would still hit the guard.

## Notifications

- @mrz1836

### DIFF
--- a/bump/bump.go
+++ b/bump/bump.go
@@ -273,29 +273,6 @@ func applyCoinbaseToSTUMP(stumpPath *transaction.MerklePath, coinbaseTxID *chain
 	}
 }
 
-// computeCorrectedSubtreeRoot parses a subtree 0 STUMP, replaces the placeholder
-// with the real coinbase txid, and returns the corrected subtree root by climbing
-// from the coinbase leaf via the BRC-74 ComputeRoot algorithm.
-//
-// The prior implementation built only the STUMP's level 0 and called
-// computeMissingHashes, then returned mp.Path[height-1][0].Hash — but that
-// element is the LEFT child of the subtree root, not the root itself
-// (the root is H(mp.Path[height-1][0], mp.Path[height-1][1])). For any
-// pow2-sized subtree ≥2 leaves, the returned value was a mid-level hash, and
-// subtreeHashes[0] was silently corrupted. The compound BUMP merge uses
-// first-seen-wins by (level, offset), so the bug only surfaced when subtree 0
-// was not the first stump processed — i.e., it depended on the Aerospike read
-// order and went undetected until a production block with 6 subtrees was
-// delivered in order [2,4,5,3,0,1].
-func computeCorrectedSubtreeRoot(stumpData []byte, coinbaseTxID *chainhash.Hash) (*chainhash.Hash, error) {
-	stumpPath, err := transaction.NewMerklePathFromBinary(stumpData)
-	if err != nil {
-		return nil, err
-	}
-	applyCoinbaseToSTUMP(stumpPath, coinbaseTxID, nil)
-	return stumpPath.ComputeRoot(coinbaseTxID)
-}
-
 // subtree0RootFromCoinbaseBUMP derives the real subtree-0 root by walking the
 // coinbase BUMP (which is a full block-level path for the coinbase tx) from
 // level 0 up through the subtree-internal levels.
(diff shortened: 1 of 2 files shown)


oskarszoon (approved):
Approve. Verified against live mainnet block 951360 (on-chain root `5ed5a7a9…`, proof depth 17 → internalHeight 17−2=15): the bounded coinbase-BUMP fold yields the true subtree-0 root `cbaaf149…` instead of over-climbing to the block root. `ceil(log2)` has no float drift in range (nice touch pinning it with `TestLog2Ceil_NoFloatDrift`), bounds are guarded, 89 tests pass.

On the `!=`→`<` relaxation: it's safe. It loosens an input-shape check, not the acceptance gate — `ValidateCompoundRoot` still recomputes against the node's own chain-header merkle root, so a malformed or over-tall STUMP can only cause a loud rejection, never a forged accepted proof. Net it strengthens vs the old unbounded climb that caused this incident.

One thing to confirm before merge, not a blocker: the coinbase-BUMP-absent path for multi-subtree blocks isn't fixed (it falls back to the STUMP height and skips the subtree-0 correction). It still fails loud — the original stuck-tx symptom — so it's no worse than today, but it's not fixed and it's untested. Does ingestion guarantee a coinbase BUMP accompanies every multi-subtree block's STUMP set? If not, worth a follow-up for that residual + a test documenting the fallback.

Optional: the safety argument rests on `ValidateCompoundRoot` staying mandatory after `BuildCompoundBUMP` — a durable invariant/test guarding that would be cheap insurance.

rafa-js:
> Approve. Verified against live mainnet block 951360 (on-chain root `5ed5a7a9…`, proof depth 17 → internalHeight 17−2=15): the bounded coinbase-BUMP fold yields the true subtree-0 root `cbaaf149…` instead of over-climbing to the block root. `ceil(log2)` has no float drift in range (nice touch pinning it with `TestLog2Ceil_NoFloatDrift`), bounds are guarded, 89 tests pass.
> 
> On the `!=`→`<` relaxation: it's safe. It loosens an input-shape check, not the acceptance gate — `ValidateCompoundRoot` still recomputes against the node's own chain-header merkle root, so a malformed or over-tall STUMP can only cause a loud rejection, never a forged accepted proof. Net it strengthens vs the old unbounded climb that caused this incident.
> 
> One thing to confirm before merge, not a blocker: the coinbase-BUMP-absent path for multi-subtree blocks isn't fixed (it falls back to the STUMP height and skips the subtree-0 correction). It still fails loud — the original stuck-tx symptom — so it's no worse than today, but it's not fixed and it's untested. Does ingestion guarantee a coinbase BUMP accompanies every multi-subtree block's STUMP set? If not, worth a follow-up for that residual + a test documenting the fallback.
> 
> Optional: the safety argument rests on `ValidateCompoundRoot` staying mandatory after `BuildCompoundBUMP` — a durable invariant/test guarding that would be cheap insurance.

Thanks for the careful review.

**Coinbase-BUMP-absent multi-subtree path:** confirmed — ingestion does *not* guarantee a coinbase BUMP. The datahub parser treats a zero-length coinbase-BUMP field as "absent" (`bump/datahub.go:378`), and `coinbaseBUMPReconciles` deliberately passes on an empty coinbase BUMP — "some peers legitimately omit it" (`bump/datahub.go:201-206`). So a multi-subtree block can reach `BuildCompoundBUMP` with `coinbaseBUMP == nil`.

In that case the subtree-0 root can't be corrected: the datahub computes it against the coinbase *placeholder*, and without the coinbase BUMP there's no real coinbase txid to fold and no full-block path to bound the climb. The compound root then can't match the header root, so the mandatory `ValidateCompoundRoot` after `BuildCompoundBUMP` (`services/bump_builder/builder.go:481`) rejects it and the tx stays non-MINED — the original stuck-tx symptom, but it **never forges an accepted proof**. So it's fail-safe and no worse than today; a correctness fix isn't possible without the coinbase data.

I've added `TestBuildCompoundBUMP_NoCoinbaseBUMP_MultiSubtree_FailsSafe` to lock that contract (compound builds, `ValidateCompoundRoot` rejects against the true block root — never silently accepts the placeholder root). I deliberately did **not** add an early-fail guard inside `BuildCompoundBUMP`: it would bypass the richer `dumpBUMPFailureInputs` diagnostic the builder emits on the validation rejection.

I'll open a separate follow-up issue to track "multi-subtree blocks without a coinbase BUMP can't be MINED" (an ingestion/datahub-guarantee question rather than a bump bug).

**`ValidateCompoundRoot` stays mandatory:** agreed it's load-bearing; the new test pins the "reject, don't forge" half of that invariant at the compound layer.

**`!=`→`<` relaxation:** agreed, thanks for confirming.

galt-tr (approved):

PR #207, opened 2026-06-18
feat(bump-builder): detect and recover silently-dropped STUMPs

freemans13:
## Summary

Consuming side of [merkle-service PR #162](https://github.com/bsv-blockchain/merkle-service/pull/162). merkle now attaches `expectedSubtreeIndices` (the set of subtree indices that produced a STUMP for each callback URL) to `BLOCK_PROCESSED`. This closes a **pre-existing silent-loss hole**: because STUMPs are sparse, arcade could not tell a lost STUMP from a legitimately-absent one, so a block missing only *some* of its STUMPs still built a compound BUMP that validated while the missing subtree's txs never reached `MINED` — no error, no metric, no recovery (the watchdog couldn't see it because `processed_at` was stamped the instant `BLOCK_PROCESSED` arrived). It also unblocks widening merkle's `callback` topic.

The fix makes **bump-builder the owner of block finalization**:

- **`handleBlockProcessed` no longer stamps `processed_at`** — it only enqueues `BLOCK_PROCESSED`. `processed_at` is the watchdog's recovery signal; stamping it on arrival is exactly what blinded the watchdog.
- **bump-builder runs a completeness gate** after the existing grace window: it compares the STUMPs it stored against `callback.ExpectedSubtreeIndices` by **set membership** (a duplicate/extra STUMP can't mask a missing one). On a gap it logs, increments `arcade_bump_builder_incomplete_stumps_total`, and returns **without** stamping `processed_at` — leaving the block in the watchdog's stale scan, which re-drives it via the **existing** block-level `POST /reprocess` (re-emitting the missing STUMPs and `BLOCK_PROCESSED`).
- **`processed_at` is now stamped only on a genuine finalize**: BUMP built and validated, a legitimate no-tracked-tx block, or the short-circuit redelivery of an already-built block.

### Why this shape

- **No new store methods, schema, config, or recovery loop.** The expected set rides the existing Kafka `BLOCK_PROCESSED` message and is re-supplied by `/reprocess` on every recovery attempt, so it never needs to be persisted in arcade. Recovery reuses the existing `services/watchdog` + `merkleservice.Client.Reprocess` (backoff, lease, recency-bounded) untouched.
- **Backward compatible.** An empty/absent expected set means "expect zero STUMPs", so an arcade running against pre-#162 merkle (field simply absent) behaves exactly as before. The only behavioural change is *where* `processed_at` is stamped (bump-builder, after grace+build) rather than *whether*.
- **Soft-fail self-heals.** If the `processed_at` stamp hits a transient store error, the block stays in the stale scan and the watchdog re-drives it; the short-circuit re-stamps on redelivery. Bounded by the watchdog's `StaleThreshold` cadence and `RecencyDepth`.

### Notes for reviewers / rollout

- Recovery of a *deferred* (incomplete) block depends on the chaintracks `UpsertBlockHeaderSeen` row supplying the real block height — this is a **pre-existing** watchdog requirement (its stale query already filters on `header_seen_at` and `block_height`), unchanged here. The old height-0 placeholder this handler used to write was never recoverable anyway.
- Observable change: `processed_at` (and the `/api/v1/blocks/processing-status` `HasBlockProcessed` flag) now flips ~grace-window later (default 30s) instead of instantly on callback receipt — it now means "finalized", not "callback received".
- Rollout order: merkle #162 **already emits** `expectedSubtreeIndices` unconditionally — there is no emit feature-flag (the `block.emitExpectedStumpSet` toggle named in merkle's design doc was never implemented; the PR's "No config flag" section rejected it). So: deploy this → arcade immediately consumes the field → verify an injected missing STUMP is detected and recovered → then widen the `callback` topic (separate change).

## Test plan

- [x] `go build ./...`, `go vet ./...` — clean
- [x] `golangci-lint run` — clean
- [x] Full suite: `go test ./...` — 637 passing
- [x] New bump-builder tests: incomplete set defers (no BUMP, no stamp); same-count index mismatch still incomplete (membership, not count); complete set finalizes + stamps; empty set + zero STUMPs finalizes; short-circuit stamps on redelivery; soft-fail self-heals via short-circuit re-stamp
- [x] New api-server tests: handler enqueues without stamping `processed_at`; Kafka enqueue failure surfaces as 500

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -270,6 +270,20 @@ var BumpBuilderEmptyStumpBlocksTotal = promauto.NewCounter(prometheus.CounterOpt
 	Help: "BLOCK_PROCESSED messages handled with zero STUMPs in the store for the block.",
 })
 
+// BumpBuilderIncompleteStumpsTotal counts BLOCK_PROCESSED messages whose
+// merkle-supplied expected-STUMP set (CallbackMessage.ExpectedSubtreeIndices)
+// was NOT fully satisfied by the STUMPs arcade had stored once the grace window
+// elapsed. Such a block is deliberately left un-finalized (processed_at stays
+// NULL) so the watchdog re-drives it via merkle's /reprocess, which re-emits the
+// missing STUMPs and BLOCK_PROCESSED. This is the metric that makes the
+// previously-silent partial-STUMP drop visible: unlike the all-missing case
+// (empty_stump_blocks_total), a block missing only SOME of its STUMPs used to
+// build a valid-looking BUMP and lose the absent subtree's txs with no signal.
+var BumpBuilderIncompleteStumpsTotal = promauto.NewCounter(prometheus.CounterOpts{
+	Name: "arcade_bump_builder_incomplete_stumps_total",
+	Help: "BLOCK_PROCESSED messages left un-finalized because expected STUMPs were still missing after the grace window.",
+})
+
 // BumpBuilderShortCircuitTotal counts BLOCK_PROCESSED messages handled by the
 // short-circuit path — the BUMP already exists in the store and this is a
 // redelivery (typically from /reprocess re-emitting BLOCK_PROCESSED). Tracks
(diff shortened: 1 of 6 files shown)


mrz1836 (approved):
TLDR; LGTM

PR #215, opened 2026-06-22
fix(validator): report unknown-parent height so post-Genesis pushes aren't capped at 520 bytes (#214)

galt-tr:
## Summary

Fixes #214. Arcade's BDK-backed validator rejected otherwise-valid modern transactions whose unlocking scripts push elements larger than 520 bytes, with `Push value size limit exceeded` — even though the same raw transactions are accepted by ARC and mined on-chain.

**Root cause.** `ValidateTransaction` validates the spending transaction at `allForksActiveHeight = 2_000_000_000` (post-Genesis/Chronicle) but reported `assumedUTXOHeight = 1` for **every** source UTXO. BDK's per-input `CalculateFlags(utxoHeight, blockHeight, consensus)` keys each input's protocol era — and therefore the historical 520-byte `MAX_SCRIPT_ELEMENT_SIZE_BEFORE_GENESIS` limit — off the **source UTXO height**, not the spending block height. Height `1` is below every network's Genesis activation height (mainnet 620538), so each input was evaluated under pre-Genesis rules and the 520-byte limit was wrongly applied to post-Genesis spends.

## Fix

Report teranode's unknown-parent sentinel (`math.MaxUint32`, the value of teranode's internal `unconfirmedParentHeight`) instead of `1`. In **policy mode** — arcade's only mode (`NewDefaultOptions` sets `SkipPolicyChecks=false`) — teranode's BDK adapter (`substituteUnconfirmedHeights`) resolves the sentinel to the post-Genesis candidate block height, so every input is evaluated against current consensus. This mirrors how teranode's own mempool intake stamps the sentinel for parents whose height it cannot resolve (`services/validator/Validator.go`).

This is the issue author's preferred fix ("adopt Teranode/BDK's candidate/unknown-parent height policy"). It is robust to future activation-height changes (no magic threshold to maintain) and semantically honest — arcade genuinely does not know parent heights at intake. arcade's path runs **no** coinbase-maturity or relative-locktime check, so the old "appears mature" rationale for height 1 has no downside.

## Tests

`TestValidate_PostGenesisLargePushes` exercises 520-, 521-, and 5210-byte pushes through the **real BDK engine**:
- **Regression guard:** arcade's public `ValidateTransaction` must accept all three sizes. This failed for 521/5210 before the fix (`utxoHeights=1 error=Push value size limit exceeded`) and passes after.
- **Root-cause control:** calling the BDK validator directly proves the rejection is purely a function of the reported source height — at pre-Genesis height `1` BDK rejects >520-byte pushes; at a post-Genesis height it accepts them. This also guards against future BDK behavior changes.

The existing `TestValidate_ChronicleVersionGate` and `TestValidate_FeeTooLow` still pass, confirming the Chronicle version-gate keys off the spending block height (not the UTXO height) and that the change is isolated.

## Verification

```
CGO_ENABLED=1 go test ./validator/...   # all pass (incl. existing regressions)
CGO_ENABLED=1 go build ./...            # clean
go vet ./validator/ + gofmt             # clean
```

### DIFF
--- a/validator/validator.go
+++ b/validator/validator.go
@@ -15,6 +15,7 @@ package validator
 import (
 	"context"
 	"fmt"
+	"math"
 
 	chaincfg "github.com/bsv-blockchain/go-chaincfg"
 	sdkTx "github.com/bsv-blockchain/go-sdk/transaction"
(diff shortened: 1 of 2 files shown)


mrz1836 (approved):
LGTM

Issue #83, opened 2026-05-01
[F-025] no-verdict broadcasts leave transactions stuck without durable retry


galt-tr:
## Summary
no-verdict broadcasts leave transactions stuck without durable retry

## Location
`services/propagation/propagator.go:299-306`

## Severity / Category / Confidence
- Severity: HIGH
- Category: Correctness
- Confidence: High

## Code
```go
			if res.status == nil {
				noVerdict++
				continue
			}
			if res.status.Status == models.StatusRejected && IsRetryableError(res.errMsg) {
				retryable++
				p.handleRetryableFailure(ctx, batch[i].TXID, res.rawTx)
				continue
```

## Issue
When all endpoints time out, are unhealthy, or return only non-final `202` responses, `res.status` is nil and the code records a metric but does not schedule durable retry. The original propagation message has already been acknowledged.

## Impact
Transactions can remain indefinitely in `RECEIVED` after an infrastructure outage or no-healthy-endpoint period, with no reaper row to rebroadcast them.

## Recommendation
Treat no-verdict outcomes as retryable infrastructure failures and call `SetPendingRetryFields` with the raw transaction, without incrementing the semantic rejection retry count if desired.

## Related cross-cutting findings
- C-002

---
*Filed from the 2026-05 third-party code review (finding `F-025`). See `REVIEW_REPORT.md` in the review bundle for full context and cross-references.*

galt-tr:
This issue has been inactive for **60 days** and will be closed in 14 days if no further activity occurs.

Ladybug7117:
We hit F-025 in production on **v0.9.0** and wanted to add a concrete reproduction — and flag that it has a **client-facing** dimension beyond the internal stuck-in-`RECEIVED` symptom.

### Reproduction

Our self-hosted Teranode's tx-submission endpoint (arcade's `teranode.broadcast_urls` target) was returning `HTTP 500` during an infrastructure degradation — the node had temporarily lost the ability to validate new tx submissions. A client POSTed a tx to arcade's broadcast API and immediately got:

```
HTTP 202   { "txStatus": "RECEIVED", "txid": "…" }
```

[...]

PR #228, opened 2026-07-08
fix(propagation): reap stale RECEIVED txs stranded by no-verdict broadcasts (#83)

galt-tr:
## Problem

A transaction could sit in `RECEIVED` **forever** after a downstream outage, presenting to the caller as a phantom broadcast (client got `202 RECEIVED` = apparent success, but the tx never reached the network). Root cause, three mechanics combining:

- **No-verdict ≠ rejected.** A broadcast that produces no verdict (downstream timeout / 5xx / no per-slot body) is classified `txResultClassRequeue` with `status == nil` and left at `RECEIVED`. It is *meant* to be retried indefinitely until the downstream heals — never rejected (marking a validated tx `REJECTED` because our own infra was down would be wrong). So there is no "retry N → REJECTED" exit for this path.
- **The only retry is non-durable.** `requeueAfterDelay` is an in-memory 2s goroutine whose sole durability is the uncommitted Kafka offset; on `ctx.Done()` it is dropped (`"requeue dropped before delay elapsed; txs left uncommitted for replay"`). A rolling upgrade during a sustained outage loses it.
- **The durable backstop skipped `RECEIVED`.** `reapOnce` only rebroadcast stale `SEEN_ON_NETWORK` / `SEEN_MULTIPLE_NODES`; `RECEIVED` hit the `default` arm and was skipped by design — its comment assumed `RECEIVED` implied an intake error the submitter owns, but the no-verdict path lands there *after* intake already returned `202` as success.

## Fix

Bring `RECEIVED` into `reapOnce`'s rebroadcast scope. A `RawTx`-carrying row stranded past a new `staleReceivedAge` (1h, mirroring `staleSeenOnNetworkAge`) is rebroadcast through the **same** `registerBatch → broadcastInChunks → applyTerminalStatuses` pipeline the SEEN path already uses.

- Recent rows (still in the normal intake/requeue path) and body-less rows (existing `len(RawTx)==0` guard) are left alone.
- Only validated txs reach `RECEIVED` — intake validation failures land at `REJECTED` — so rebroadcasting is always safe and correct.
- This preserves the intended **indefinite self-heal** of a transient downstream outage rather than introducing a false give-up-to-`REJECTED`.

Also reconciles the now-inaccurate `reapOnce` comments and adds the reaper's **first unit tests**.

## Scope

This addresses the **durable-retry half** of #83 (the reporter's option 1). The separate client-facing concern — `202 RECEIVED` presented to the caller as terminal success (option 2) — is intentionally out of scope and tracked separately.

## Testing

- `services/propagation/` new tests: `TestReapOnce_RebroadcastsStaleReceived` (the #83 repro: register count 0→1) and `TestReapOnce_ReceivedSelection` (stale-with-body rebroadcast; recent / body-less / terminal skipped; stale SEEN still works). ✅
- Full `go test ./services/propagation/` ✅
- `golangci-lint run ./services/propagation/...` → 0 issues ✅
- `go build ./...` ✅

Refs #83


mrz1836 (approved):
LGTM

Ladybug7117:
Thanks @galt-tr — we're the ones who hit this on v0.9.0 and added the repro above. Pulled the branch and tested it; here's what we found.

**red→green.** Our standalone reaper test fails on base `main`: a validated tx stranded at `RECEIVED` still carrying its `RawTx`, aged past the stale window, gets `register count = 0` while a stale `SEEN_ON_NETWORK` row still rebroadcasts — so it isolates the `RECEIVED` gap. On this branch it passes, as do your `TestReapOnce_RebroadcastsStaleReceived` / `TestReapOnce_ReceivedSelection`. `go test ./services/propagation/`, `go vet`, and `gofmt` are clean here.

**Approach.** Reads right to us: only validated txs reach `RECEIVED` (validation failures land at `REJECTED`), so rebroadcast is safe; the `len(RawTx)==0` guard and `staleReceivedAge` (mirroring `staleSeenOnNetworkAge`) cover the leave-alone cases — body-less rows, and rows still inside the normal intake / 2s in-memory requeue window. Reusing `registerBatch → broadcastInChunks → applyTerminalStatuses` keeps it consistent with the SEEN path.

**Scope.** Agreed `202` is the right accept state and shouldn't change — the reaper skipping these txs was the defect. The caller-side reading of `202` (treating it as terminal success) is the separate piece we're handling on our end.

The durable-retry gap reads as closed against our testing. Appreciate the quick turnaround.

PR #229, opened 2026-07-08
perf(app): gate validator + tx-tracker hydration to modes that use them

galt-tr:
## Problem
Only **api-server** consumes `deps.Validator` and `deps.TxTracker` (see `BuildServices`), yet `Bootstrap` built both for **every** mode. Two startup costs therefore hit every arcade pod regardless of mode:

- `validator.NewValidatorForNetwork` builds **two** cgo/BDK script-verification engines (`app/app.go:159`).
- `TxTracker.LoadFromStore` streams the **entire** transactions history into an in-memory map (`app/app.go:205`); with chaintracks disabled it passes `currentHeight=0`, so nothing is pruned and the whole table loads.

On a busy cluster this full-table hydration scales with tx volume and **OOMKilled every 512Mi arcade-v2 component** (`sse`, `watchdog`, `p2p-client`, `bump-builder`) ~18s into startup — before the health port even binds. A quiet cluster (17–42 MiB/pod) masked it entirely. This is what forced the interim 2Gi bump in [bsva-infra-flux#230](https://github.com/bsv-blockchain/bsva-infra-flux/pull/230).

## Fix
Gate both behind `modeNeedsValidator` / `modeNeedsTxTracker`, mirroring the existing `modeNeedsChaintracks` pattern in the same file. Only `api-server` / `all` pay the cost now; the other modes start in tens of MiB and their limits can return to 512Mi.

Also corrects a stale comment claiming `bump-builder` relies on tracker hydration — its tracker prefilter was removed; only api-server reads the tracker today.

## Not in this PR (follow-up)
`store.IterateStatusesSince` (the hydration query, tracker-only) over-fetches `raw_tx` and other blobs it discards. A projection-only variant (`SELECT txid,status,block_height`) would cut api-server's own hydration allocation ~10-100×. Kept separate since it touches the store interface + postgres/aerospike impls and deserves its own review/benchmark.

## Testing
- `go build ./app/`, `go vet ./app/` clean.
- `go test ./app/... ./store/...` pass.
- Added `TestModeNeedsTxTracker` / `TestModeNeedsValidator` (table tests mirroring `TestModeNeedsChaintracks`).

https://claude.ai/code/session_0152bwyqKUBRhMkHcD5LMXsz


mrz1836 (approved):
LGTM

PR #299, opened 2026-08-11
feat(propagation): multi-partition arcade.propagation with family keying and a missing-parent safety net

galt-tr:
Closes #295.

## What this does

Lifts the single-partition intake ceiling (~1.6–1.9k TPS measured in #295) by letting `arcade.propagation` run N partitions with one dep-aware dispatcher per partition claim — without ever letting an out-of-order broadcast produce a false terminal REJECTED.

## The ordering model (replacing the #151 exact-1 rule)

Two mechanisms replace total topic order:

1. **Family partition keys at intake.** `handleSubmitTransactions` runs a union-find over each submitted batch: all txs connected through in-batch spends share one partition key (the lexicographically smallest member txid — order-independent, so shuffled resubmissions map to the same partition). The headline case — 2 parents + 98 children in one POST — lands whole on one partition, so per-partition Kafka order still delivers parents to their dispatcher before children, and Teranode's *no parent and child in one `/txs` batch* contract holds structurally. Isolated txs keep today's txid key.
2. **Missing-parent safety net.** Chains that span submissions (child POSTed after its parent, possibly on another partition/pod) have no ordering guarantee — the child can reach Teranode first and draw `TX_MISSING_PARENT` (or `TX_NOT_FOUND`, the un-wrapped store-layer form). Per the #254 principle these are **conditions, not verdicts**: the tx rides the existing bounded requeue, parks at `PENDING_RETRY` on budget exhaustion (reaper owns durable retry — the no-wedge invariant from #294 is preserved), and the *only* path to REJECTED is arcade's own store showing an ancestor terminally REJECTED (the cross-partition analog of the dispatcher's same-partition cascade, with the same `parent rejected (ancestor …)` reason). This also fixes a latent false-REJECT: the reaper's dep-blind rebroadcast of a child whose parent was still parked previously terminalized the child REJECTED.

Enforcement moves from `CheckExactPartitions(…, 1)` to a fail-closed `CheckMinPartitions` against new config `propagation.partitions` (default 1 — existing deployments upgrade with zero action; a missing topic still aborts startup).

## Multi-dispatcher correctness + the serial-path hot spot

- **Per-claim `dispatcherIO` channel sets.** The dispatcher's terminal/requeue channels lived on the shared `Propagator`; with two claims, partition B's loop could consume partition A's terminal event, leaving A's offset never `Done` and its commit watermark frozen — the #294 wedge class, silently reintroduced. Each claim now gets its own channel set, threaded through the batch pipeline; the reaper's terminal notifications (nil io) fan out to every registered dispatcher (at most one owns the tx, the rest no-op).
- **Delta-accounted depth gauges.** `arcade_propagation_pending_depth` / `inflight_depth` (and the atomics behind the reaper's backlog log line) become pod totals across partition dispatchers, with each loop's contribution removed on claim teardown. No metric schema change.
- **`advanceMarks` amortization (#295's suspected 10x).** The full `pendingMarks` map walk per terminal event is replaced by an ordered `markQueue` popped on the existing 50 ms flush tick (plus a teardown drain): amortized O(1) per message instead of O(terminals × in-flight). Claim-revocation semantics unchanged — only offsets strictly below `LowestUnfinished` are ever marked.
- **Producer ordering hardening.** The sync producer is now idempotent with `MaxOpenRequests=1` (protocol 2.6): a broker-side retry can no longer reorder a family's messages within a partition. Note: this bumps the wire protocol for all sync produces — fine for Redpanda/modern Kafka.
- **`retry_backoff_ms` is now actually read** (it was declared-but-dead config). Its viper default moves 500→2000 so shipped cadence is unchanged; deployments that explicitly set it will start feeling their configured value.

## Testing

- **Smoke, real topology in-process**: the memory broker gains sarama-compatible key-hash partitions and per-partition claims. The 10k-tx chained-batch ordering test now also runs at 4 partitions — all four invariants (no cohabitation, parent-precedes-child, exactly-once, chunk cap) hold; sabotaging the intake back to txid keying makes it fail with dozens of child-before-parent inversions, so the guard is load-bearing. New smoke tests drive the out-of-order child (submitted before its parent, 4 partitions) recovering to ACCEPTED via missing-parent retry with zero REJECTED, and an orphan child parking at PENDING_RETRY while unrelated traffic keeps flowing.
- **Unit**: family-key table (star/chain/fan-in/reorder-stability), min-partition contract, two-claim terminal routing isolation, reaper nil-io fan-out, markQueue, missing-parent line matcher (wrapped/deepest-cause/TX_NOT_FOUND vs conflict/policy negatives), verdict-vs-condition precedence (peer verdict wins, sticky accept wins), store-consult cascade both ways, budget exhaustion parks-not-rejects, reaper leaves-row/cascades regression pair.
- `-race` clean on `services/propagation` and `kafka`; `golangci-lint` clean.

## Observability

New `arcade_propagation_missing_parent_total{outcome="requeued"|"rejected_ancestor"}`. A burst of `requeued` is expected during a partition widen (key remap window); a sustained steady-state rate means family keying isn't co-locating same-submission chains. Dashboards note: PENDING_RETRY baseline may rise (children of slow parents); worth checking the stuck-transient alert tolerates it, and considering PENDING_RETRY in the #291 census as a follow-up.

## Rollout

Topic-first, per the "Widening an existing deployment" section in docs/production-kafka.md: `rpk topic add-partitions` → raise `propagation.partitions` → roll pods → scale replicas up to the partition count. During the remap window watch `missing_parent_total{outcome="requeued"}` (expected burst) and `outcome_total{outcome="rejected"}` (a spike would mean ordering artifacts reached a verdict — the thing this PR makes impossible by construction). Zero-race alternative: drain and recreate (the #151 cutover pattern).

Follow-ups deliberately out of scope: a multi-partition e2e against real redpanda (smoke covers the topology in-process; the e2e harness currently has the podman registry pull issue), and standby right-sizing on the scale cluster.

---

## Follow-up: PENDING_RETRY convergence (review response)

An adversarial review of this branch asked how long the `PENDING_RETRY` path
takes to resolve for a worst-case parent/child topology. The answer at the
time was: too long, and sometimes never.

The safety net this PR leans on — `parkExhaustedRequeues`, the reaper's
`PENDING_RETRY` arm, `stalePendingRetryAge`, `staleScanLookback` — all
pre-existed on `main`, where it was a rare poison-batch escape hatch. This PR
promotes it to the designed steady-state landing spot for every
cross-submission chain, and it did not have the properties that role needs.

### What was wrong

At shipped defaults a parked chain of depth D drained in
`10s + 300s + (D-6) x 30s` — about 52 minutes at depth 100 — because
`reapOnce` built its messages with `InputTXIDs` empty and broadcast the whole
selection as a single `/txs` call, so exactly one dependency level could clear
per tick. That also put parents and children in one batch, the shape the
design says teranode forbids.

Three cases never converged at all:

- **Starvation.** Selection was `IterateStatusesSince`, which Postgres serves
  `ORDER BY timestamp_at DESC` and Pebble ascending, capped per tick. The
  missing-parent arm deliberately left the row alone and
  `MarkMerkleRegisteredByTxIDs` only writes `merkle_registered_at`, so a
  failing row's `timestamp_at` was frozen at park time. On Postgres it sank
  below every newer eligible row; on Pebble a block of unresolvable orphans
  sat at the head and starved the walk's other arms deployment-wide.
- **Silent abandonment.** At `park + 24h` a frozen row fell out of
  `staleScanLookback` and stayed `PENDING_RETRY` permanently — no terminal
  status, no log line, no metric, and `GET /tx` never reached a terminal
  answer.
- **Parent that never arrives.** `rejectedAncestor` returns false on
  `ErrNotFound`, so a tx spending a non-existent outpoint — the most common
  client mistake there is — parked immortally, where on `main` it was
  REJECTED in about a second.

### What changed

**Drain by `next_retry_at`, not by scanning `timestamp_at`.** The store
already had a complete durable retry queue — `retry_count`, `next_retry_at`,
`idx_tx_retry_ready(next_retry_at) WHERE status='PENDING_RETRY'`, and
`BumpRetryCount` / `SetPendingRetryFields` / `GetReadyRetries` (`ORDER BY
next_retry_at LIMIT n FOR UPDATE SKIP LOCKED`) / `ClearRetryState` implemented
on all three backends — with zero callers outside `store/`. Wiring the park
path to it fixes starvation, the frozen timestamps, the Postgres/Pebble
ordering divergence and multi-process safety in one change.

**Broadcast one dependency layer at a time.** `layerByDependency` splits the
selected batch so each layer's accepts are visible to the next, which also
removes the parent-and-child-in-one-batch shape.

**Give up visibly.** After `pending_retry_max_attempts` (default 288, about
24h at the backoff cap — roughly when rows used to be abandoned) an
unresolvable tx is terminalized REJECTED with a reason the submitter can read.

**Walk ancestors transitively.** `rejectedAncestor` consulted direct parents
only while documented as the analog of `cascadeReject`'s recursive BFS, so a
REJECTED grandparent behind a `PENDING_RETRY` parent cost one retry cycle per
generation to condemn.

**Family keys are computed over the full submitted batch.** The key is
`min(family)`, so it moved whenever a member left the set — and the intake
dedup CAS drops members constantly. `POST {P}` then `POST {P, C}` deduped the
parent out and re-keyed the child onto a partition its in-flight parent was
not on, losing the dispatcher's dep-hold entirely.

### Numbers

| case | before | after |
|---|---|---|
| depth-5 parked chain | 5 reaper ticks | 1 tick |
| depth-500 parked chain, 200-row cap | ~4h, or never | **3 ticks (~90s)** |
| depth-12 chain, reverse order, 12 requests, 4 partitions (smoke, compressed timers) | did not converge | **169ms** |
| orphan whose parent never arrives | PENDING_RETRY forever, silently dropped at 24h | REJECTED with a reason |

### Observability

New `arcade_propagation_parked_depth` and
`arcade_propagation_parked_oldest_age_seconds` (PENDING_RETRY is absent from
`stuckTransientStatuses` and `reaper_ready_depth` saturates at the per-tick
cap, so the parked set was unbounded and unmeasurable), plus
`arcade_propagation_pending_retry_total{scheduled,exhausted}`. The reaper path
now reports `missing_parent_total{outcome="reaper_retry"}` instead of
`"requeued"` — a steady population of unresolvable orphans previously pinned
that counter at a constant rate, which reads exactly like the family-keying
regression its own dashboard note tells operators to watch for. All outcome
children are pre-registered so alerts get 0 rather than no-data.

### Also in this follow-up

- `docs/production-kafka.md` is rewritten for multi-partition. It still
  documented `arcade.propagation` as "exactly 1 (mandatory)" and referenced
  `CheckExactPartitions`, a function this PR deletes; an operator following it
  would create a 1-partition topic and crash-loop any pod configured for more.
  The "Widening an existing deployment" section now exists.
- `app.go` keeps `CheckPartitions`' nil topic list. Passing it the real
  fan-out topics turned a check that has been dead since `TopicTransaction`
  was retired into a live boot-blocking one: a deployment with a leftover
  `kafka.min_partitions` above its actual `tx_status` / `block_processed`
  width would crash-loop on upgrade, for an unrelated reason.
- `propagation` block added to `config.example.yaml`;
  `PROPAGATION_PARTITIONS` wired through `docker-compose.yaml` to `topic-init`,
  which read the variable but was never passed it.

### New config

`propagation.pending_retry_backoff_ms` (5000),
`pending_retry_max_backoff_ms` (300000 — the old flat eligibility age, so the
slowest cadence is unchanged while early attempts are far faster),
`pending_retry_max_attempts` (288).

### Testing

- Six pinning tests landed failing first, then fixed: convergence, fairness,
  liveness, durability, the durable happy path, and the ancestor walk depth.
- `layerByDependency` across chain / fan-out / fan-in / diamond /
  out-of-batch parents / self-reference / cycle guard.
- End-to-end smoke for a deep chain submitted in reverse across separate
  requests on 4 partitions. This caught a bug in the fix itself:
  cascade-parked descendants were being written with a NULL `next_retry_at`
  and so were invisible to the new drain.
- `mockStore.GetReadyRetries` now honors the real query's status filter and
  `next_retry_at` ordering, with insertion-order tie-breaks; ties are the
  common case and were previously served from a Go map in random order,
  masking how much convergence depends on that ordering.
- Full unit suite, `-race` on `services/propagation`, `kafka`,
  `services/api_server`, `-tags=smoke`, `-tags=postgres`, and
  `golangci-lint` all clean.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -100,8 +100,14 @@ func Bootstrap(ctx context.Context, cfg *config.Config, logger *zap.Logger) (*De
 	// dependency family, so per-partition order still delivers a parent's
 	// admit to its dispatcher before its children's, and cross-partition
 	// stragglers are absorbed by the missing-parent safety net instead of
-	// being terminally REJECTED. `>=` rather than `==` because extra
-	// partitions only cost idle dispatchers.
+	// being terminally REJECTED.
+	//
+	// `>=` rather than `==` because a wider topic is safe to run against —
+	// not because the extra partitions are inert. Nothing passes this value
+	// to a partitioner, so the producer hashes over the topic's ACTUAL
+	// partition count: every partition carries real traffic and gets its own
+	// dispatcher. This knob is a fail-closed assertion, not a placement
+	// control; lowering it does not narrow the producer's hash space.
 	if pErr := kafka.CheckMinPartitions(broker, kafka.TopicPropagation, cfg.Propagation.Partitions, logger); pErr != nil {
 		_ = producer.Close()
 		return nil, nil, fmt.Errorf("kafka partition check: %w", pErr)
@@ -110,9 +116,16 @@ func Bootstrap(ctx context.Context, cfg *config.Config, logger *zap.Logger) (*De
 	// are currently no other hot-path topics that need it (TopicTransaction
 	// was retired with tx_validator) but the knob is retained so a future
 	// fan-out topic can opt into the check without re-introducing config.
+	//
+	// Deliberately still passed a nil topic list, which makes CheckPartitions
+	// a no-op. Handing it the real fan-out topics turns a check that has been
+	// dead since TopicTransaction was retired into a live boot-blocking one:
+	// any deployment carrying a leftover kafka.min_partitions greater than
+	// its actual arcade.tx_status / arcade.block_processed width would
+	// crash-loop on upgrade, for a reason unrelated to multi-partition
+	// propagation. Enabling it is its own change, with its own release note.
 	if cfg.Kafka.MinPartitions > 1 {
-		topics := []string{kafka.TopicStatusUpdate, kafka.TopicBlockProcessed}
-		if pErr := kafka.CheckPartitions(broker, topics, cfg.Kafka.MinPartitions, logger); pErr != nil {
+		if pErr := kafka.CheckPartitions(broker, nil, cfg.Kafka.MinPartitions, logger); pErr != nil {
 			_ = producer.Close()
 			return nil, nil, fmt.Errorf("kafka partition check: %w", pErr)
 		}
--- a/app/app.go
+++ b/app/app.go
@@ -82,19 +82,27 @@ func Bootstrap(ctx context.Context, cfg *config.Config, logger *zap.Logger) (*De
 		zap.String("store_backend", cfg.Store.Backend),
 	)
 
-	broker, err := kafka.NewBroker(cfg.Kafka)
+	// The memory backend provisions its own topics, so it takes the
+	// propagation partition count directly; the sarama backend ignores the
+	// map (real topics are broker-provisioned and checked below).
+	broker, err := kafka.NewBroker(cfg.Kafka, map[string]int{
+		kafka.TopicPropagation: cfg.Propagation.Partitions,
+	})
 	if err != nil {
 		return nil, nil, fmt.Errorf("creating kafka broker: %w", err)
 	}
 	producer := kafka.NewProducer(broker)
 
-	// Hard requirement: TopicPropagation MUST be single-partition. The
-	// dep-aware dispatcher's single-goroutine state ownership relies on
-	// total order at the topic level — parent/child txids on different
-	// partitions would bypass the in-memory dep index and reintroduce
-	// the cross-batch "missing inputs" race we just removed. Check this
-	// on every startup regardless of MinPartitions.
-	if pErr := kafka.CheckExactPartitions(broker, kafka.TopicPropagation, 1, logger); pErr != nil {
+	// TopicPropagation must exist with at least propagation.partitions
+	// partitions (fail-closed on a missing topic — auto-creation would let
+	// the broker default decide the key→partition mapping). Since #295 the
+	// topic may have any partition count: the api server keys messages by
+	// dependency family, so per-partition order still delivers a parent's
+	// admit to its dispatcher before its children's, and cross-partition
+	// stragglers are absorbed by the missing-parent safety net instead of
+	// being terminally REJECTED. `>=` rather than `==` because extra
+	// partitions only cost idle dispatchers.
+	if pErr := kafka.CheckMinPartitions(broker, kafka.TopicPropagation, cfg.Propagation.Partitions, logger); pErr != nil {
 		_ = producer.Close()
 		return nil, nil, fmt.Errorf("kafka partition check: %w", pErr)
 	}
@@ -103,7 +111,8 @@ func Bootstrap(ctx context.Context, cfg *config.Config, logger *zap.Logger) (*De
 	// was retired with tx_validator) but the knob is retained so a future
 	// fan-out topic can opt into the check without re-introducing config.
 	if cfg.Kafka.MinPartitions > 1 {
-		if pErr := kafka.CheckPartitions(broker, nil, cfg.Kafka.MinPartitions, logger); pErr != nil {
+		topics := []string{kafka.TopicStatusUpdate, kafka.TopicBlockProcessed}
+		if pErr := kafka.CheckPartitions(broker, topics, cfg.Kafka.MinPartitions, logger); pErr != nil {
 			_ = producer.Close()
 			return nil, nil, fmt.Errorf("kafka partition check: %w", pErr)
 		}
--- a/services/api_server/handlers.go
+++ b/services/api_server/handlers.go
@@ -1176,15 +1176,27 @@ func (s *Server) handleSubmitTransactions(c *gin.Context) {
 	// Phase 4: build propagation envelopes and publish as one batch.
 	// input_txids drives the dispatcher's dep-aware admission — children
 	// of any in-flight parent are held until the parent terminalizes.
+	// Messages are keyed by dependency family (#295): txs of this
+	// submission connected through in-batch spends share a key, so on a
+	// multi-partition topic a whole family lands on one partition and
+	// per-partition order still delivers parents before children.
 	if len(toPublish) > 0 {
+		txids := make([]string, len(toPublish))
+		inputs := make([][]string, len(toPublish))
+		for i, p := range toPublish {
+			txids[i] = p.txid
+			inputs[i] = collectInputTXIDs(p.tx)
+		}
+		keys := familyPartitionKeys(txids, inputs)
+
 		msgs := make([]kafka.KeyValue, 0, len(toPublish))
-		for _, p := range toPublish {
+		for i, p := range toPublish {
 			msgs = append(msgs, kafka.KeyValue{
-				Key: p.txid,
+				Key: keys[i],
 				Value: map[string]interface{}{
 					"txid":        p.txid,
 					"raw_tx":      p.raw,
-					"input_txids": collectInputTXIDs(p.tx),
+					"input_txids": inputs[i],
 				},
 			})
 		}
--- a/services/api_server/handlers.go
+++ b/services/api_server/handlers.go
@@ -1028,11 +1028,6 @@ func (s *Server) handleSubmitTransactions(c *gin.Context) {
 	// Phase 1: parse the whole stream. Use the canonical TxID() (not a
 	// hash of the wire bytes) so Extended Format submissions key the
 	// same as the canonical txid the propagator broadcasts.
-	type parsedItem struct {
-		tx   *sdkTx.Transaction
-		raw  []byte
-		txid string
-	}
 	var parsed []parsedItem
 	offset := 0
 	for offset < len(body) {
@@ -1181,57 +1176,8 @@ func (s *Server) handleSubmitTransactions(c *gin.Context) {
 	// multi-partition topic a whole family lands on one partition and
 	// per-partition order still delivers parents before children.
 	if len(toPublish) > 0 {
-		// Keys are computed over `parsed` — the FULL submitted batch — not
-		// over toPublish. The distinction is the difference between a stable
-		// family key and a moving one.
-		//
-		// The key is min(family), so it moves whenever a member leaves the
-		// set. Members leave constantly: the Phase 3 dedup CAS drops every tx
-		// already known at a non-REJECTED status. The standard wallet pattern
-		// POST {P} then POST {P, C} therefore dedups P out, leaving the child
-		// with no in-batch parent and keying it by its own txid — onto a
-		// partition its still-in-flight parent is not on, where the
-		// dispatcher cannot hold it behind that parent. It draws
-		// TX_MISSING_PARENT and parks, where before #295's multi-partition
-		// topic it was admitted in milliseconds. The partial-produce retry
-		// path has the same shape: the members that already landed dedup out
-		// and the remainder recomputes onto a different partition.
-		//
-		// A deduped-out member can remain the family representative because
-		// the key is only a hash input; it does not have to be a txid that
-		// this request publishes.
-		allTxids := make([]string, len(parsed))
-		allInputs := make([][]string, len(parsed))
-		for i, p := range parsed {
-			allTxids[i] = p.txid
-			allInputs[i] = collectInputTXIDs(p.tx)
-		}
-		allKeys := familyPartitionKeys(allTxids, allInputs)
-		keyByTxid := make(map[string]string, len(parsed))
-		inputsByTxid := make(map[string][]string, len(parsed))
-		for i, p := range parsed {
-			if _, ok := keyByTxid[p.txid]; ok {
-				continue // duplicate txid in one body: first occurrence wins
-			}
-			keyByTxid[p.txid] = allKeys[i]
-			inputsByTxid[p.txid] = allInputs[i]
-		}
-
-		msgs := make([]kafka.KeyValue, 0, len(toPublish))
-		for _, p := range toPublish {
-			key, ok := keyByTxid[p.txid]
-			if !ok {
-				key = p.txid
-			}
-			msgs = append(msgs, kafka.KeyValue{
-				Key: key,
-				Value: map[string]interface{}{
-					"txid":        p.txid,
-					"raw_tx":      p.raw,
-					"input_txids": inputsByTxid[p.txid],
-				},
-			})
-		}
+		keyByTxid, inputsByTxid := familyKeysBySubmittedTxid(parsed)
+		msgs := propagationMessages(toPublish, keyByTxid, inputsByTxid)
 		if err := s.producer.SendBatch(ctx, kafka.TopicPropagation, msgs); err != nil {
 			if errors.Is(err, kafka.ErrBrokerBackpressure) {
 				s.logger.Warn("batch submit rejected: kafka backpressure", zap.Int("count", len(msgs)))
--- a/services/api_server/handlers.go
+++ b/services/api_server/handlers.go
@@ -1181,22 +1181,54 @@ func (s *Server) handleSubmitTransactions(c *gin.Context) {
 	// multi-partition topic a whole family lands on one partition and
 	// per-partition order still delivers parents before children.
 	if len(toPublish) > 0 {
-		txids := make([]string, len(toPublish))
-		inputs := make([][]string, len(toPublish))
-		for i, p := range toPublish {
-			txids[i] = p.txid
-			inputs[i] = collectInputTXIDs(p.tx)
+		// Keys are computed over `parsed` — the FULL submitted batch — not
+		// over toPublish. The distinction is the difference between a stable
+		// family key and a moving one.
+		//
+		// The key is min(family), so it moves whenever a member leaves the
+		// set. Members leave constantly: the Phase 3 dedup CAS drops every tx
+		// already known at a non-REJECTED status. The standard wallet pattern
+		// POST {P} then POST {P, C} therefore dedups P out, leaving the child
+		// with no in-batch parent and keying it by its own txid — onto a
+		// partition its still-in-flight parent is not on, where the
+		// dispatcher cannot hold it behind that parent. It draws
+		// TX_MISSING_PARENT and parks, where before #295's multi-partition
+		// topic it was admitted in milliseconds. The partial-produce retry
+		// path has the same shape: the members that already landed dedup out
+		// and the remainder recomputes onto a different partition.
+		//
+		// A deduped-out member can remain the family representative because
+		// the key is only a hash input; it does not have to be a txid that
+		// this request publishes.
+		allTxids := make([]string, len(parsed))
+		allInputs := make([][]string, len(parsed))
+		for i, p := range parsed {
+			allTxids[i] = p.txid
+			allInputs[i] = collectInputTXIDs(p.tx)
 		}
-		keys := familyPartitionKeys(txids, inputs)
+		allKeys := familyPartitionKeys(allTxids, allInputs)
+		keyByTxid := make(map[string]string, len(parsed))
+		inputsByTxid := make(map[string][]string, len(parsed))
+		for i, p := range parsed {
+			if _, ok := keyByTxid[p.txid]; ok {
+				continue // duplicate txid in one body: first occurrence wins
+			}
+			keyByTxid[p.txid] = allKeys[i]
+			inputsByTxid[p.txid] = allInputs[i]
+		}
 
 		msgs := make([]kafka.KeyValue, 0, len(toPublish))
-		for i, p := range toPublish {
+		for _, p := range toPublish {
+			key, ok := keyByTxid[p.txid]
+			if !ok {
+				key = p.txid
+			}
 			msgs = append(msgs, kafka.KeyValue{
-				Key: keys[i],
+				Key: key,
 				Value: map[string]interface{}{
 					"txid":        p.txid,
 					"raw_tx":      p.raw,
-					"input_txids": inputs[i],
+					"input_txids": inputsByTxid[p.txid],
 				},
 			})
 		}
--- a/services/api_server/partition_keys.go
+++ b/services/api_server/partition_keys.go
@@ -1,0 +1,98 @@
+package api_server
+
+// familyPartitionKeys returns the Kafka partition key for each tx of a
+// submitted batch. Txs connected through in-batch spends — tx i lists an
+// input whose source txid is also in the batch — form a dependency family
+// and share one key: the lexicographically smallest txid of the family.
+// Isolated txs key by their own txid (keys[i] == txids[i]), matching the
+// pre-#295 behavior and the single-submit path.
+//
+// Same key ⇒ same partition ⇒ per-partition Kafka order delivers every
+// parent's admit to its dispatcher before its children's, which is what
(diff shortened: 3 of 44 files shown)


galt-tr (commented):
Adversarial pass focused on one question: **for a worst-case topology — batches with dense parent/child relationships, submitted out of order — how long does the `PENDING_RETRY` path take to converge, and can it fail to converge at all?**

First, what I checked and found correct, so it's not in doubt: per-claim `dispatcherIO` isolation, the `markQueue` watermark logic (monotone delivery ⇒ amortized O(1), never over-commits), `newSyncProducerConfig`'s idempotence + `MaxOpenRequests=1`, and `partitionForKey` vs. sarama's `hashPartitioner` — the arithmetic is mod-then-abs in both, identical expression. The mechanism work is sound.

The concern is architectural, and it's this:

> `parkExhaustedRequeues`, `stalePendingRetryAge`, `staleScanLookback` and the reaper's `PENDING_RETRY` arm **all pre-exist on `main`**, where they are a *rare poison-batch escape hatch*. This PR promotes that path to the **designed steady-state landing spot for every cross-submission chain**. It does not have the properties that role requires.

## Convergence, at shipped defaults

`retry_max_attempts=5`, `retry_backoff_ms=2000`, `reaper_interval_ms=30000`, `reaper_rebroadcast_batch=200`, `stalePendingRetryAge=5m` (a package const, not config).

```
fast path  = 1 initial + 5 requeues × 2s   ≈  10s   → absorbs depth ≤ 6
park floor = stalePendingRetryAge          = 300s
per level  = reaper_interval_ms            =  30s   ← because the reaper is dep-blind

T(D) ≈ 10s + 300s + (D−6) × 30s
```

| chain depth D | wall clock to full ACCEPTED |
|---|---|
| 6 | ~10 s (never parks) |
| 10 | ~7 min |
| 50 | ~27 min |
| 100 | **~52 min** |
| 200 | ~1 h 42 min |
| > 200 | **may never converge** |

The `(D−6) × 30s` term exists because `reapOnce` builds `propagationMsg{TXID, RawTx}` with **empty `InputTXIDs`**, does no topological sort, and puts all 200 rows in a single Teranode `/txs` call. At most one dependency level can clear per tick.

## Three ways it does not converge

1. **Frozen timestamps + newest-first scan.** The reaper's missing-parent arm deliberately leaves the row alone, and `MarkMerkleRegisteredByTxIDs` only writes `merkle_registered_at` — so a failing row's `timestamp_at` is pinned at park time. `IterateStatusesSince` is `ORDER BY timestamp_at DESC`, early-stopped at `rebroadcastBatch`. Old parked rows sink below every newer eligible row and are never revisited. All rows in one park batch also share a single `now`, so the tie-break order is arbitrary but *stable* — if D > 200 and the resolvable frontier isn't in the selected 200, progress is zero, permanently.

2. **Silent abandonment at 24 h.** Once a frozen row ages past `staleScanLookback`, `WHERE timestamp_at >= $1` excludes it. It stays `PENDING_RETRY` **forever** — no terminal status, no log line, no metric. That is currently the only exit.

3. **Parent that never arrives.** `rejectedAncestor` returns `false` on `store.ErrNotFound`, so a tx spending a genuinely non-existent outpoint — the most common client mistake there is — parks immortally, where on `main` it was terminally REJECTED in about a second. 200 of those consume the entire per-tick window and stop the reaper's *other* duties deployment-wide (RECEIVED self-heal, ACCEPTED→SEEN nudge, SEEN refresh), while re-POSTing 200 × *E endpoints* of doomed transactions every 30 s indefinitely.

## The fix is already in the tree, unused

`retry_count`, `next_retry_at`, `idx_tx_retry_ready(next_retry_at) WHERE status='PENDING_RETRY'`, and `BumpRetryCount` / `SetPendingRetryFields` / `GetReadyRetries` (`ORDER BY next_retry_at LIMIT n FOR UPDATE SKIP LOCKED`) / `ClearRetryState` are all implemented on the `Store` interface with **zero callers outside `store/` and its tests**. `parkExhaustedRequeues` writes neither field.

Routing the park through it fixes the LIFO starvation, the frozen timestamps, the flat 5-minute floor, the Postgres-`DESC`-vs-Pebble-ascending divergence, and multi-pod reaper safety in one change — and lights up an index that already exists.

## The topology in question also defeats the PR's throughput goal

`familyPartitionKeys` merges components across a *single* spend edge, so one consolidation/fan-in tx unions an entire batch into one family → one key → one partition → one dispatcher goroutine. With `maxBatchBytes` at 256 MiB this is unbounded, and there's no family-size cap and no key-skew metric. Worth noting the new multi-partition smoke test would pass unchanged if every key collapsed onto partition 0 — it asserts the four ordering invariants but never that more than one partition carried traffic.

Related: `POST /tx` (`handlers.go:853`) still keys by bare txid and is untouched. Under #151's single partition, parent-then-child via two sequential single submits was totally ordered and the dispatcher held the child. That guarantee is gone with nothing replacing it, so every one-tx-per-request client now lands on the missing-parent path by default.

## Smaller items

- **`docs/production-kafka.md` isn't in the diff.** The PR body cites "the new *Widening an existing deployment* section" — that section doesn't exist, and the shipped doc still says `arcade.propagation` must be **exactly 1** partition, shows `--partitions 1` examples, and references `CheckExactPartitions`, which this PR deletes. An operator following it creates a 1-partition topic and any pod with `propagation.partitions > 1` crash-loops.
- **Nothing in the repo asserts per-partition ordering** — the foundation the whole model rests on. `_OneClaimPerPartition` records only message keys and discards the payload sequence. `TestNewSyncProducerConfig_OrderingInvariants` also doesn't assert `Producer.Partitioner` is the hash partitioner; swapping it for round-robin destroys every guarantee here with no test going red.
- `arcade_propagation_missing_parent_total{outcome="requeued"}` is incremented on the reaper path too, where nothing is requeued — which poisons exactly the signal the PR's dashboard guidance tells operators to watch. Neither label child is pre-registered, so alerts return no-data rather than 0.
- `PENDING_RETRY` isn't in `stuckTransientStatuses`, so the stuck census can't see it, and `reaper_ready_depth` saturates at 200. The parked set is unbounded and unmeasurable — while this PR makes it an expected steady state.

Detail on each is in the inline comments. Happy to take the fixes and the pinning tests.

services/propagation/reaper.go
```diff
@@ -436,6 +437,18 @@ func (p *Propagator) reapOnce(ctx context.Context) {
...
 			}
+		case txResultClassMissingParent:
+			// The reaper's dep-blind rebroadcast can resurface a child
+			// before its parent (both parked, or the parent wedged on
+			// another pod). Pre-#295 this drew TX_MISSING_PARENT and
+			// falsely terminalized the child REJECTED. Condition, not
+			// verdict: only a REJECTED ancestor in arcade's own store
+			// condemns the child; otherwise the row stays as-is and the
+			// next tick retries after the parent lands.
+			if cascaded, _ := p.resolveMissingParents(ctx, []propagationMsg{registered[i]}, []string{res.errMsg}); len(cascaded) > 0 {
```
galt-tr: **This is the O(D) term in the convergence math.** `reapOnce` builds its messages as `propagationMsg{TXID: st.TxID, RawTx: st.RawTx}` — `InputTXIDs` is left empty — and then hands the whole selected batch to `broadcastInChunks` as one `/txs` call with no topological layering.

So the reaper cannot resolve more than **one dependency level per tick**, because a child can only succeed once its parent is in Teranode's UTXO set, and every level goes out simultaneously. At `reaper_interval_ms=30000` a parked chain of depth D costs `(D−6) × 30s` on top of the 5-minute park floor: ~52 minutes at D=100.

Two consequences beyond the latency:

1. Because `InputTXIDs` is empty, `parentTxIDsOf` falls through to `sdkTx.NewTransactionFromBytes(msg.RawTx)` — a **full transaction parse for every row on every tick**, forever, for rows that never resolve.
2. All 200 rows in one `/txs` call is exactly the parent-and-child-in-one-batch shape the design says Teranode forbids.

Suggested fix: populate `InputTXIDs` here (parse once), then layer the batch by in-batch dependency and emit one chunk per layer sequentially within the tick. That collapses D ticks into 1 — the difference between ~52 minutes and a few seconds.

services/propagation/reaper.go
```diff
@@ -436,6 +437,18 @@ func (p *Propagator) reapOnce(ctx context.Context) {
...
+				rejected += len(cascaded)
+				terminalStatuses = append(terminalStatuses, cascaded...)
+			}
 		case txResultClassUnknown, txResultClassRequeue:
 			// Requeue / Unknown from the reaper's rebroadcast path:
```
galt-tr: **"Leave the row alone" is what makes the retry set starve and then silently vanish.**

Nothing on this path writes the row, and `MarkMerkleRegisteredByTxIDs` only sets `merkle_registered_at` — so a failing row's `timestamp_at` stays **frozen at park time**. Combined with the selection query, which is:

```sql
SELECT ... FROM transactions WHERE timestamp_at >= $1 ORDER BY timestamp_at DESC
```

early-stopped once `len(stuck) >= p.rebroadcastBatch` (200), that gives two failure modes:

- **Starvation.** Any steady supply of newer eligible rows keeps the newest-200 window permanently ahead of older parked rows, which are never revisited. And since `parkExhaustedRequeues` stamps one `now := time.Now()` for the whole batch, a large park lands as a block of identical timestamps whose tie-break order is arbitrary but *stable* — if the resolvable frontier isn't in the chosen 200, progress is zero forever.
- **Silent abandonment.** At `park + staleScanLookback` (24 h) the row falls out of the `WHERE` clause entirely. It stays `PENDING_RETRY` **permanently** — no terminal status, no log, no metric, and `GET /tx` never reaches a terminal answer for the client.

Note the Pebble path walks `idx:tx:updated:` ascending, so the two backends starve in *opposite* directions — worth a test that pins both.

services/propagation/propagator.go
```diff
@@ -642,6 +670,88 @@ func cascadeReason(status models.Status, ancestor string) string {
...
+//
+// Deliberately conservative: store.ErrNotFound, read errors, and every
+// non-REJECTED status (RECEIVED, PENDING_RETRY, DOUBLE_SPEND_ATTEMPTED,
+// …) return false — anything else would invent a verdict. Runs on
+// processBatch / reaper worker goroutines only, never the dispatcher.
+func (p *Propagator) rejectedAncestor(ctx context.Context, msg propagationMsg) (string, bool) {
+	if p.store == nil {
+		return "", false
+	}
+	for _, parent := range parentTxIDsOf(msg) {
```
galt-tr: **This walks direct parents only — one hop — but the reason string and the comment both say "ancestor".**

`cascadeReason` renders `parent rejected (ancestor %s)`, and the doc comment above calls this "the cross-partition analog of the dispatcher's same-partition cascade" — but `cascadeReject` in `dispatcher.go` is a *recursive* BFS over the waiter graph, and this is a single `GetStatus` per direct parent.

Concretely: grandparent REJECTED, parent at `PENDING_RETRY`, child asking. The child is not condemned — it requeues, waits for the parent to cascade on a later tick, and only then cascades itself. So even the *rejection* path is O(depth) reaper ticks at 30 s each, when it could be one store walk.

Worth a bounded transitive walk with a visited set, plus per-batch memoization so a fan-in tx doesn't re-read shared ancestors once per parent.

Two smaller things in the same function: `err != nil` and `row == nil` are conflated, so a store blip reads identically to "parent is fine" (safe direction, but it means a Postgres hiccup mass-parks); and `ErrNotFound` — the never-submitted parent — is the branch that produces the immortal `PENDING_RETRY` rows.

services/propagation/propagator.go
```diff
@@ -642,6 +670,88 @@ func cascadeReason(status models.Status, ancestor string) string {
...
+			)
+			cascaded = append(cascaded, &models.TransactionStatus{
+				TxID:      msg.TXID,
+				Status:    models.StatusRejected,
+				Timestamp: now,
+				ExtraInfo: cascadeReason(models.StatusRejected, ancestor),
+			})
+			continue
+		}
+		metrics.PropagationMissingParentTotal.WithLabelValues("requeued").Inc()
```
galt-tr: **This counter also fires on the reaper path, where nothing is requeued.**

`reapOnce` calls `resolveMissingParents(...)` and discards the `requeue` return (`if cascaded, _ := ...`), then explicitly leaves the row for the next tick. But this line increments unconditionally, so every reaper tick over a parked row counts as a "requeue".

That matters because the PR body tells operators:

> A burst of `requeued` is expected during a partition widen … a sustained steady-state rate means family keying isn't co-locating same-submission chains.

With K immortal orphans this counter sits at a constant `K/reaper_interval` forever, which permanently poisons the exact signal being introduced to diagnose family-keying regressions. Suggest a distinct `outcome="reaper_retry"`.

Also: neither label child is in `metrics/preregister.go`, so both series are absent until first use and `rate()`/threshold alerts return no-data instead of 0 — the same gap `PreRegisterTxSubmissions` exists to close for the other vectors.

services/api_server/handlers.go
```diff
@@ -1176,15 +1176,27 @@ func (s *Server) handleSubmitTransactions(c *gin.Context) {
...
+	// multi-partition topic a whole family lands on one partition and
+	// per-partition order still delivers parents before children.
 	if len(toPublish) > 0 {
+		txids := make([]string, len(toPublish))
+		inputs := make([][]string, len(toPublish))
+		for i, p := range toPublish {
+			txids[i] = p.txid
+			inputs[i] = collectInputTXIDs(p.tx)
+		}
+		keys := familyPartitionKeys(txids, inputs)
```
galt-tr: **The family key is computed over `toPublish` — the post-dedup slice — which makes it unstable across submissions.**

`partition_keys.go`'s doc comment claims "a client retrying the same family with the batch shuffled maps to the same partition as the original messages." That holds for shuffling, but not for the far more common case: a member dropping out via the Phase 3 dedup CAS.

Take the standard wallet pattern — `POST {P}`, then moments later `POST {P, C}` (re-sending the parent for safety):

1. P dedups out at a non-REJECTED status, so `toPublish = {C}`.
2. `familyPartitionKeys([C], [[P]])` finds no in-batch source, so **C is keyed by its own txid**.
3. C lands on a partition that isn't P's. That dispatcher's `inFlight` has never seen P, so `handleAdmit` does not hold it.
4. C broadcasts concurrently with P, draws `TX_MISSING_PARENT`, burns 5 × 2 s, parks — and waits at least 5 more minutes for the reaper.

Pre-#299 this was held cleanly and accepted in milliseconds. Same shape on the partial-produce path: `SendMessages` fails for a subset, the client retries the identical body, the succeeded members dedup out, and the remainder recomputes to a different key than the members already in flight.

Fix looks cheap: compute the keys over `parsed` (the full submitted batch) and look each survivor's key up by txid. The key is only a hash input — it needn't be a published txid — so a deduped-out member remaining as the representative is exactly what restores stability.

services/api_server/partition_keys.go
```diff
@@ -0,0 +1,97 @@
...
+		parent[rb] = ra
+		size[ra] += size[rb]
+	}
+
+	for i, ins := range inputs {
+		for _, src := range ins {
+			if src == "" || src == txids[i] {
+				continue
+			}
+			if j, ok := index[src]; ok {
```
galt-tr, line 77: **One edge merges two components, so a single transaction can collapse an entire batch onto one partition.**

The union-find itself is right — near-linear, order-independent, path-halving, and `TestFamilyPartitionKeys_OrderIndependent` pins the property. The issue is what "family" means for real traffic.

A 10,000-tx POST of mutually independent transactions plus **one** consolidation tx spending an output of each yields exactly one component → one key → one partition → one dispatcher goroutine, i.e. precisely the 1.6–1.9k TPS serial ceiling #295 exists to lift. `maxBatchBytes` is 256 MiB, so there's no practical bound on how large that family gets, and there's no family-size cap and no key-skew metric to see it happening.

This isn't hypothetical for the workload — consolidation and sweep transactions are exactly the fan-in shape, and a chained load generator producing one long chain per batch collapses the same way.

Two things would help a lot: a family-size histogram / per-partition publish counter so skew is visible, and a note in the config docs that a single-family batch serializes by design. Also worth knowing that `BuildChains` in the smoke suite is linear-only, so the *merge* branch exercised here is never tested above the 5-element unit tables.

app/app.go
```diff
@@ -103,7 +111,8 @@ func Bootstrap(ctx context.Context, cfg *config.Config, logger *zap.Logger) (*De
 	// was retired with tx_validator) but the knob is retained so a future
 	// fan-out topic can opt into the check without re-introducing config.
 	if cfg.Kafka.MinPartitions > 1 {
-		if pErr := kafka.CheckPartitions(broker, nil, cfg.Kafka.MinPartitions, logger); pErr != nil {
+		topics := []string{kafka.TopicStatusUpdate, kafka.TopicBlockProcessed}
```
galt-tr: **This converts a dead check into a live boot-blocking one, and it's unrelated to the PR's subject.**

On `main` the call is `kafka.CheckPartitions(broker, nil, cfg.Kafka.MinPartitions, logger)` — and `CheckPartitions` ranges over that topics slice, so passing `nil` made it a **complete no-op**. Anyone who set `kafka.min_partitions` has been running with it silently doing nothing.

With a real topics list it now hard-fails startup. Concretely: an existing deployment with `kafka.min_partitions: 4` and `arcade.tx_status` provisioned at 2 partitions upgrades to this PR and gets

```
kafka partition check: topic arcade.tx_status has 2 partitions, need at least 4
```

and crash-loops — for a reason that has nothing to do with multi-partition propagation and isn't mentioned in the PR body or the rollout section.

Either restore `nil` and do this in its own PR, or keep it and call it out explicitly in the rollout notes as an upgrade precondition.

app/app.go
```diff
@@ -82,19 +82,27 @@ func Bootstrap(ctx context.Context, cfg *config.Config, logger *zap.Logger) (*De
...
-	if pErr := kafka.CheckExactPartitions(broker, kafka.TopicPropagation, 1, logger); pErr != nil {
+	// TopicPropagation must exist with at least propagation.partitions
+	// partitions (fail-closed on a missing topic — auto-creation would let
+	// the broker default decide the key→partition mapping). Since #295 the
+	// topic may have any partition count: the api server keys messages by
+	// dependency family, so per-partition order still delivers a parent's
+	// admit to its dispatcher before its children's, and cross-partition
+	// stragglers are absorbed by the missing-parent safety net instead of
+	// being terminally REJECTED. `>=` rather than `==` because extra
+	// partitions only cost idle dispatchers.
```
galt-tr: Minor, but this comment will mislead an operator: extra partitions do **not** "only cost idle dispatchers."

Nothing passes `cfg.Propagation.Partitions` to a partitioner — sarama's `hashPartitioner` gets `numPartitions` from cluster metadata, so the producer hashes over the **actual** topic partition count. If the topic has 8 and config says 4, all 8 carry real traffic and the consumer group assigns all 8. `propagation.partitions` is purely a fail-closed startup assertion; it has no effect on placement.

Worth saying that directly here, since the natural reading is that lowering the knob narrows the producer's hash space.

PR #309, opened 2026-08-15
fix(propagation): never terminalize a transaction without a reason a submitter can act on

galt-tr:
## What Changed

A review of every rejection Teranode can produce, and what survives the trip to a wallet-visible row. Four fixes plus the matrix and a contract test that keeps it honest.

- **`teranode/client.go` keeps the verdicts that keyed.** `parseTxsFailures` returns `TxsFailures{ByKey, Unkeyed}` instead of `nil` for the whole body when one line carries no txid.
- **`propagator.go` narrows.** A chunk holding a line arcade cannot place is re-broadcast in halves until the line is the only answer to a single transaction, where "named nobody" can only mean "named this one". Bounded at ~2·log2(n) extra requests, on the failure path only.
- **Node faults stop being verdicts.** `STORAGE_ERROR` / `SERVICE_ERROR` / `ERROR`, and opaque `PROCESSING` lines whose peer answered 5xx, requeue instead of writing a permanent `REJECTED`. An opaque line with a **422** is read as the missing-parent condition; one with a **403** carries the frozen-utxo class. The status is consulted only when the peer reported a single failure, where it describes that transaction exactly.
- **The give-up reason quotes what the network last said** instead of telling every exhausted transaction that a parent it spends never arrived.

Plus `docs/teranode-error-surfacing.md` (the full matrix) and `services/propagation/teranode_reason_contract_test.go`.

## Why It Was Necessary

Production showed **4,357** `REJECTED` rows whose entire reason was `PROCESSING (4): [ProcessTransaction][<txid>] failed to validate transaction`, and a further class of failures that never terminalized at all — they requeued until the durable-retry budget ran out, then terminalized claiming a parent had never reached the network, whatever had actually happened.

The cause is a boundary we don't own. Teranode's `errors.UserMessage` surfaces only the deepest cause on its public allowlist and discards the rest — **including the `[ProcessTransaction][<txid>]` wrapper that is our only reliable way to bind a verdict to a transaction**. So for whole classes of rejection the line names no transaction:

```
TX_INVALID (31): bad-txns-in-belowout
TX_POLICY (39): transaction fee is too low
TX_INVALID (31): GoBDK fail to ValidateTransaction
TX_LOCK_TIME (35): lock time (…) is not less than median block time (…)
```

`parseTxsFailures` failed closed on exactly that shape: one txid-less line discarded the **whole body**. The verdicts that did key were thrown away, every transaction in the chunk requeued — including ones the peer had accepted — and the failing one looped until `schedulePendingRetry` gave up with a reason unrelated to the real failure. That rule was written for a rare Teranode panic-recovery accident; it turned out to describe the common case.

Two other classes were misfiled rather than lost: node-side faults (`STORAGE_ERROR`, block-assembly unavailable) became permanent verdicts, and missing-parent — invisible because `TX_MISSING_PARENT` is not allowlisted — terminalized every out-of-order child as `REJECTED` and cascaded that onto its descendants, defeating the #295/#299 safety net entirely.

## Testing Performed

New `teranode_reason_contract_test.go`: every failure class Teranode can produce, run at chunk cap 1 **and** 1024, asserting the verdict binds to the right transaction, lands in the right class, and carries a cause-specific reason — with the siblings' acceptances intact.

Its fixtures are **generated by Teranode's own error constructors rendered through its own `UserMessage`**, never hand-written. This matters: the pre-existing fixtures in this package all carry a `[ProcessTransaction]` wrapper that Teranode strips whenever an allowlisted cause is present, so the suite was green on strings Teranode never emits. **27 of 34 subtests failed** against the old behaviour; all pass now. Because arcade already imports the Teranode module, a version bump that changes the boundary fails these tests instead of changing production behaviour silently.

Also covered: narrowing settles a 16-transaction chunk with one guilty member inside the bisection bound and condemns only that one; a fully-keyed response never enters the narrowing path; `giveUpReason` for all three cases.

`go test ./...` passes except `logfields.TestNoCanonicalKeyLiteralsOutsidePackage`, which fails identically on an unmodified tree here (it scans stale `.claude/worktrees` copies). `golangci-lint run` reports 0 issues on `services/propagation/...` and `teranode/...`.

## Impact / Risk

**One deliberate relaxation of #292.** Implicit accepts are now granted when *every* failure line has been placed. Withholding them was right while an alien line meant "some unidentifiable transaction failed"; once outpoint attribution accounts for every line that uncertainty is gone, and Teranode's body carries one line per failed transaction, so absence really does mean acceptance. The withholding still applies whenever any line is left unplaced. This is the change most worth a second opinion.

**Five existing tests encoded contracts this changes** — each rewritten with the reasoning, and two now use content-aware stubs so they model a node that answers about the transactions it was actually sent, rather than condemning whatever it is given. `TestBroadcast_422RejectionSurfacesExtraInfo` moved its verdict peer from 422 to 400: 422 was a mislabel, since it is reachable only from `ErrTxMissingParent`.

**Narrowing costs extra round trips on the failure path**, worst case ~2·log2(chunk size). It never runs when the response is fully attributable, which is the steady state, and re-submitting an accepted transaction is idempotent (`ErrTxExists` → success).

**Three gaps remain and can only be closed upstream** — BDK script causes, `TX_MISSING_PARENT`, `UTXO_FROZEN`. bsv-blockchain/teranode#1595 does that. Each is pinned here by an `upstreamGap` row with a `mustNotContain` token, so when Teranode starts carrying the real cause the row fails and asks to be promoted rather than going stale. Note the coupling: after that PR merges and we bump past `v0.16.0-beta-9`, `TestBoundaryLine_DropsTheProcessTransactionWrapper` will fail **by design** to say the 422 fallback has become redundant.

Follows #292, #294, #299, #302.

## Notifications
- @galt-tr

https://claude.ai/code/session_01K8pDVtifV8xhKfhEya8wcJ


galt-tr:
Rebased onto `main` now that #313 has landed. Force-push replaced a `Merge branch 'main'` commit that had brought in pre-#313 main; that merge was a clean auto-merge carrying no hand-resolved content, and everything it pulled in is an ancestor of the new base, so nothing was lost.

Four conflicts, all in the area #313 touched. How they were resolved:

**Kept #313's primitive over mine.** Both PRs had grown a `lineIsOpaqueProcessing`. Mine only inspected the leading code, so a chain-leaking line — `PROCESSING (4): … TX_INVALID (31): …` — read as opaque. `lineHasNestedTeranodeVerdict` correctly reads that as a verdict whatever wrapper sits in front of it, which is also how `lineIsMissingParentCondition` already behaves. Mine is deleted.

**Kept this branch's routing**, which refines #313 where evidence exists. Teranode derives its HTTP status per transaction: 422 only from `ErrTxMissingParent`, 403 only from a frozen utxo, 5xx only from its own faults. So an opaque line whose peer reported a *single* failure gets classified rather than blanket-requeued — and in particular a 400 is a permanent verdict (`INVALID_ARGUMENT` / `COINBASE_IMMATURE` / `UTXO_INVALID_SIZE`, none allowlisted, so none can surface by name) that must not loop until the retry budget dies.

**Adopted #313's rule where that evidence is absent.** When several failures share one body the status is an aggregate describing none of them. This branch previously rejected there — producing exactly the reasonless `REJECTED` row the PR exists to remove. It now requeues, and terminates through the durable-retry budget with the line quoted in the give-up reason.

I considered narrowing the chunk to disambiguate instead, since the bisection machinery is already here, and rejected it: the shape that produces many opaque lines at once is a node-wide fault, and bisecting a 50-tx chunk into singletons would send ~100 requests to a peer that is already failing. Requeue costs one.

@GenericCPU's `TestLineIsOpaqueProcessing` and `TestBroadcast_OpaqueProcessing_RequeuesInsteadOfRejects` pass against this routing unmodified. Added `TestBroadcast_OpaqueProcessing_AmbiguousStatus_RequeuesAlongsideRealVerdict` for the case above, and checked it fails without the rule (the opaque tx picks up a `REJECTED` row whose `ExtraInfo` is the bare `PROCESSING (4)` line — the production shape).

`golangci-lint run ./services/propagation/... ./teranode/...` reports 0 issues; `go test ./...` is clean apart from the `logfields` scan that trips on stale local `.claude/worktrees` copies and passes in a clean checkout.

PR #311, opened 2026-08-16
fix(intake): do not persist REJECTED on BDK PROCESSING errors

GenericCPU:
## What Changed
- Detect teranode `ERR_PROCESSING` (BDK/cgo engine failure) with `validator.IsInternalEngineError`.
- Intake now returns **HTTP 503** + `Retry-After` and **does not** write a REJECTED row or fire REJECTED SSE/webhooks.
- Script/fee/policy failures still persist REJECTED + 400.

## Why It Was Necessary
`validator/errmap.go` already says PROCESSING/cgo is not attributable to the tx, then intake treated **every** `ValidateTransaction` error as a terminal byte-verdict. A valid transaction could land `REJECTED` because the C++ engine hiccuped.

## Testing Performed
- `go test ./validator ./services/api_server -count=1`
- New: mapped PROCESSING still unwraps; fee errors are not 503; stubbed intake 503 leaves store/Kafka empty.

## Impact / Risk
- Callers must retry 503 (same as Kafka backpressure). Do not treat 503 as a permanent reject.
- Real script failures remain 400 REJECTED.

## Notifications
- Related to opaque Teranode `PROCESSING (4)` on the propagator (separate PR).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/services/api_server/handlers.go
+++ b/services/api_server/handlers.go
@@ -26,8 +26,16 @@ import (
 	"github.com/bsv-blockchain/arcade/models"
 	"github.com/bsv-blockchain/arcade/telemetry"
 	"github.com/bsv-blockchain/arcade/teranode"
+	"github.com/bsv-blockchain/arcade/validator"
 	"github.com/bsv-blockchain/arcade/version"
 )
+
+// intakeValidator is the submit-path validator. *validator.Validator
+// implements it; tests stub it when they need an engine failure that must
+// not persist REJECTED.
+type intakeValidator interface {
+	ValidateTransaction(ctx context.Context, tx *sdkTx.Transaction, skipFees bool) error
+}
 
 // collectInputTXIDs returns the parent txids referenced by every input of
 // tx. Empty for coinbase. Used to populate the propagation envelope's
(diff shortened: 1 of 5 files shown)


galt-tr (changes requested):
This PR doesn't do what you think it does; the generic processing error comes from teranode which I am fixing here: https://github.com/bsv-blockchain/teranode/pull/1595

If we see this error from teranode it is genuinely rejected; observing all logs in production all PROCESSING error logs come directly from teranode never due to an unmapped error in our internal validator.

I don't think this change improves any observed behavior yet; this really seems like extra defensive mechanism against something I've never observed yet... which is that our validator returns something unmapped. I'd rather fix that than defer to a 503

galt-tr:
I'm going to close this and if you feel strongly I am incorrect on my take please let me know; hoping the other PRs improve this error response significantly

GenericCPU:
No problem at all — thanks for clarifying. I was mainly trying to surface a possible edge case around how `ERR_PROCESSING` gets classified, not push a required change. Happy to leave this closed and follow the Teranode fix / other PRs on the error path.

PR #313, opened 2026-08-16
fix(propagation): requeue opaque Teranode PROCESSING instead of REJECTED

GenericCPU:
## What Changed
- Per-slot Teranode lines that are only `PROCESSING (4)` (no nested `TX_*` / `UTXO_*` / other named code) no longer go into `rejectionLine`.
- Those slots fall through to `txResultClassRequeue`, matching the comment that already treated PROCESSING as infra, not a terminal verdict.
- Nested wrappers (`PROCESSING (4): … TX_INVALID (31)`) still REJECTED.
- `TestBroadcast_422RejectionSurfacesExtraInfo` now pins a real `TX_INVALID` ExtraInfo line rather than opaque PROCESSING.

## Why It Was Necessary
Teranode wraps many failures in `PROCESSING (4)`. When the line has **no nested validator/UTXO code**, it is the catch-all (engine/infra), not a byte verdict. The propagator still wrote `REJECTED` for that shape, so a valid tx could be terminalized because a peer said PROCESSING.

Intake BDK `ERR_PROCESSING` is a separate change: https://github.com/bsv-blockchain/arcade/pull/311

## Testing Performed
- `go test ./services/propagation -count=1`
- New: `TestLineIsOpaqueProcessing`, `TestBroadcast_OpaqueProcessing_RequeuesInsteadOfRejects` (no REJECTED write; delayed requeue lands).

## Impact / Risk
- Genuine `TX_INVALID` / `UTXO_SPENT` / conflict lines still REJECTED.
- Opaque PROCESSING will retry until budget/reaper, same as body-less 5xx. Callers should not treat that as a permanent reject.

## Notifications
- Related to https://github.com/bsv-blockchain/arcade/pull/311 (intake PROCESSING → 503).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/services/propagation/classify_test.go
+++ b/services/propagation/classify_test.go
@@ -7,10 +7,10 @@ import (
 
 // TestClassifyFailureLine pins the conservative Teranode-line → ARC-code map
 // (issue #254 / external feedback item 2). Invariants: the verbatim Teranode
-// line is always preserved in the message (the caller keeps every per-tx line
-// terminally rejected — the requeue signal is a body-less 5xx, decided
-// elsewhere); only confidently-mappable code prefixes gain an ARC status
-// code; and the non-final family carries an explicit retryable hint.
+// line is always preserved in the message; only confidently-mappable code
+// prefixes gain an ARC status code; and the non-final family carries an
+// explicit retryable hint. Opaque PROCESSING is routed to requeue by the
+// broadcast loop (lineIsOpaqueProcessing) and does not reach this helper.
 func TestClassifyFailureLine(t *testing.T) {
 	cases := []struct {
 		name          string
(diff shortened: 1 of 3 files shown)


galt-tr:
I have a feeling this overlaps heaviyl with https://github.com/bsv-blockchain/arcade/pull/309; checking

galt-tr:
Merging this first; and then rebasing 309 on top

galt-tr (approved):

galt-tr:
Pushed `b7c1fd2` to this branch to clear the `magex lint` failure — the PR has `maintainer_can_modify` on, shout if you would rather have taken it yourself.

Two findings, both from the deliberate no-op arm:

```
propagator.go:2681:1  `if failed` has complex nested blocks (complexity: 4) (nestif)
propagator.go:2689:44 empty-block: this block is empty, you can remove it (revive)
```

`revive`'s `empty-block` rule objects to the `else if lineIsOpaqueProcessing(line) {}` branch even though the comment inside explains it, and the extra arm pushed the enclosing `if failed` block to nesting complexity 4 — exactly the configured `nestif` threshold.

Negating the guard covers both:

```go
if lineIsMissingParentCondition(line) {
    missingParentLine[j] = preferRejectionLine(missingParentLine[j], line)
} else if !lineIsOpaqueProcessing(line) {
    rejectionLine[j] = preferRejectionLine(rejectionLine[j], line)
}
```

That is the same shape your attributed-line loop already uses a few lines below, so the two call sites now read alike. No behaviour change: an opaque PROCESSING line still leaves `rejectionLine` empty and falls through to `txResultClassRequeue`, and both `TestBroadcast_OpaqueProcessing_RequeuesInsteadOfRejects` and `TestLineIsOpaqueProcessing` still pass. Verified locally with `golangci-lint run ./services/propagation/...` (0 issues) and `go test ./services/propagation -count=1`.

Unrelated: the `Analyze (go)` failure on the previous head was a GitHub-side outage ("Encountered an error while trying to determine feature enablement: HttpError: No server is currently available"), not anything in the diff — the push should re-run it clean.

Issue #212, opened 2026-06-19
[Feature] GET /policy endpoint not implemented — clients cannot discover fee and size policy


arkadiuszpx:
### Is your feature request related to a problem?

**Labels:** `missing-feature`, `api-compatibility`

**Arcade was evaluated as a drop-in replacement for ARC in a production broadcasting pipeline.**

```markdown
## Summary

Arcade does not implement the `GET /v1/policy` endpoint defined in the ARC API contract.
Clients that call this endpoint before constructing or submitting transactions receive
HTTP 404 with no equivalent alternative.

## Environment

- Repository: `github.com/bsv-blockchain/arcade`
- Commit: `a937b6aa06a394ffe54d32e1e868a4f582f9f377`
- Reference spec: ARC (`github.com/bitcoin-sv/arc`)

## Expected behaviour (ARC contract)

```sh
GET /v1/policy
```

```json
{
  "policy": {
    "miningFee": {"satoshis": 1, "bytes": 1000},
    "maxtxsizepolicy": 10000000,
    "maxscriptsizepolicy": 500000,
    "maxtxsigopscountspolicy": 4294967295,
    "standardFormatSupported": true
  },
  "timestamp": "2026-06-10T12:00:00Z"
}
```

## Actual behaviour (Arcade)

```sh
curl https://arcade-us-1.bsvb.tech/policy
# HTTP 404 — no equivalent endpoint exists
```

## Impact

**Blocking.** Clients that call `GET /v1/policy` to discover the current fee rate or enforce
size constraints before building transactions receive HTTP 404 and cannot proceed. There is
no alternative endpoint to substitute.

### Proposed change

Implement `GET /policy` returning Arcade's current fee and size policy in the ARC schema:

- `policy.miningFee` — fee rate as `{ satoshis, bytes }`
- `policy.maxtxsizepolicy` — maximum accepted transaction size in bytes
- `policy.maxscriptsizepolicy` — maximum script size in bytes
- `policy.maxtxsigopscountspolicy` — maximum signature operations per transaction
- `policy.standardFormatSupported` — whether standard (non-EF) format is accepted
- `timestamp` — policy timestamp (RFC 3339)

```

PR #314, opened 2026-08-17
feat(policy): track network max tx/script size from node_status, not just fee

galt-tr:
Follow-up to #274 / #212.

## Why

#274 made arcade's **mining fee** dynamic: the p2p-client records the fee each teranode advertises in its `node_status` `fee_policy`, and an api-server refresher pushes the lowest fresh observation into the validator — so intake enforces, and `GET /policy` reports, the cheapest fee any node on the network will accept.

Teranode advertises `maxscriptsizepolicy` and `maxtxsizepolicy` in that **same** `fee_policy` object ([teranode `bb7e4f04`](https://github.com/bsv-blockchain/teranode/commit/bb7e4f049a362a8a27ab72d2c45ecf4b628e4478), teranode#943), but `recordPeerFee` was reading only `FeePolicy.MiningFee` and throwing both away. Arcade advertised a hardcoded 10 MiB / 500 KB regardless of what the network actually enforced.

This wires them through, so an oversized transaction is rejected at intake instead of being forwarded to a teranode that will refuse it. **No teranode change is needed** — the data has been on the wire since #943.

## Aggregation: most permissive wins

The size limits take the **maximum** across fresh peers. That is the size-limit reading of exactly the rule the fee already uses:

| | rule | meaning |
|---|---|---|
| fee | `min` across peers | accept if **at least one** node would mine it |
| size limits | `max` across peers | pre-reject only if **no** node would accept it |

Taking the minimum instead would let one node running the 10 MiB default — or one peer advertising `maxtxsize=1000` — shrink the limit for everyone, rejecting transactions the rest of the network mines happily.

## Safety

`node_status` is unauthenticated gossip, and `tnvalidator.NewTxValidator` **panics** rather than erroring on an out-of-range BDK setting (`ScriptVerifierGoBDK.go:171,183`). A discovered value is bounded at both ends before it can reach an engine:

- **floor** at arcade's built-in default, in `discoveredLimit` — observation can only ever *widen* a limit, never narrow it, so a hostile or misconfigured peer cannot cause false rejections;
- **ceiling** at the BDK consensus limit (and script size at the effective tx size), in `resolveMaxTxSize` / `resolveMaxScriptSize`, which every engine-construction path goes through;
- the rebuild is wrapped in a `recover()` that retains the previous policy, so even a clamp bug cannot take an api-server pod down.

## Config

`validator.max_tx_size_policy` and `validator.max_script_size_policy` now default to **0**, the same "not operator-configured" sentinel `min_fee_per_kb` uses. A non-zero value pins the limit and the refresher leaves that dimension alone; each dimension is pinned independently.

`max_tx_sigops_counts_policy` stays config-only — `0` already means "unlimited" there, so it cannot double as an unset sentinel.

> **Operator note:** this is a behaviour change for anyone relying on the old non-zero defaults. Unset now means "track the network" rather than "10 MiB / 500 KB forever". Until a peer is heard, or if every peer is stale, the built-in defaults still apply, so a deployment with no p2p peers behaves exactly as before.

## Validator: one atomic snapshot

The fee, the size limits and both BDK engines now live in a single immutable `effectivePolicy` behind one `atomic.Pointer`. Previously the fee was an atomic while the size accessors read plain struct fields — with sizes dynamic, that would let `GET /policy` advertise a limit the engine was not enforcing.

The fee-agnostic engine (the `X-SkipTxValidation` path) is rebuilt too, since it enforces the same size limits — but only when a size actually changed, so a fee-only update still costs one rebuild. `SetMinFeePerKB` remains as a thin wrapper.

## Store

`PeerPolicy` gains the two limits across all three backends. **Zero means "peer did not advertise"** — a teranode predating `fee_policy`, or a value too large to store — and readers skip it rather than reading it as a real limit. An unstorable size is zeroed rather than failing the write, so a peer cannot remove itself from the fee minimum by sending garbage in another field.

Postgres gets the two columns plus an idempotent `ALTER ... ADD COLUMN IF NOT EXISTS` for live deployments; pebble and aerospike need no migration since a missing JSON key / bin already reads as 0.

## Also

Corrects four stale rows in the README config table, which named `validator.max_tx_size` / `max_script_size` / `max_sig_ops` — keys that do not exist — and defaults that no longer applied.

## Verification

\`\`\`
go build ./... && go vet ./...          # clean
go test ./... && go test -race ./...    # all pass
go test -tags=postgres ./store/postgres # pass (embedded PG, exercises the schema change)
golangci-lint run                       # 0 issues
\`\`\`

New coverage: `highestObservedLimit` and `discoveredLimit` tables (max over fresh, stale excluded, zero-means-unadvertised, hostile peer cannot narrow below the default); `SetNetworkPolicy` clamping (consensus ceiling, `MaxUint64` saturation, script capped at tx size) and no-op-keeps-engines; per-dimension pinning; a `HostilePeerCannotBreakIntake` end-to-end guard; `recordPeerPolicy` size capture, legacy-peer zeros, and unstorable-size-keeps-fee; store round-trips including a row written without the new fields; and a guard that the config defaults and validator defaults have not drifted apart.


galt-tr:
CI note: the single test failure here (`TestProbe_HealthySuccess_DoesNotResetBroadcastCounter`, `teranode` package) is a pre-existing flake, not a regression from this PR — this branch changes no file under `teranode/`.

The test records every failure needed to trip the health breaker while a 5 ms probe loop runs, then immediately asserts the endpoint is unhealthy. A probe success against an unhealthy endpoint legitimately restores it (`recordProbeSuccess` treats one good probe as full recovery), so the assertion races the ticker and loses on a loaded runner. Reproduced deterministically with a 20 ms sleep before the assertion.

Fix is in #317 (test-only). Once that merges I'll rebase this branch and CI should go green.

PR #319, opened 2026-08-18
fix(policy): stop one peer zeroing the network fee floor

galt-tr:
## The report

`arcade-v2-us-1` and `arcade-v2-eu-1` (v0.13.2) both advertise **`miningFee: 0` sat/kB** on `GET /policy`, while every endpoint in `GET /health` advertises **100**:

```
$ curl .../policy   → "miningFee": {"satoshis": 0, "bytes": 1000}
$ curl .../health   → 6 endpoints at 100 sat/kB, 2 stale with no policy, none at 0
```

`maxtxsizepolicy` is the discovered `100000000` rather than the 10 MiB default, so discovery was working and the peer rows were fresh. The validator really was holding a 0 floor.

### 1. The observed fee had no lower bound

The fee takes the **minimum** across fresh peers, so a single peer advertising 0 set the floor to 0 for the whole instance. Peer announcements are unauthenticated gossip, and a node need not even be malicious to send it — teranode advertises `min_mining_tx_fee=0` when its policy settings are nil.

The size limits already had exactly this guard, and the reasoning was written down when it was added in #314:

```go
// sizes — floored, so observation can only ever widen from the built-in default
if !ok || observed < floor { return floor }

// fee — no floor at all
if !found || perKB < best { best = perKB }
```

It simply was not carried across to the fee. Now it is: discovery may track the cheapest node down to **1 sat/kB** but never to 0. Accepting zero-fee transactions is a deliberate operator decision, so **`accept_zero_fee` is the only route to a 0 floor**, and it still works — there is a test for that.

### 2. The fee counted peers arcade cannot broadcast to

`recordPeerPolicy` ran **ahead of** the datahub-URL gate, so a peer advertising no URL — or one rejected by SSRF validation, which this network really does produce (nodes announcing cluster-internal names) — set the network minimum while appearing nowhere in `/health`.

That is exactly why the report was unexplainable from outside: the operator could see only peers advertising 100.

The ordering was deliberate when the fee was introduced (#212): a peer exposing a fee but no usable URL still counted toward the minimum. This reverses it. The policy is recorded once the URL is in the registry, so `/policy`'s inputs are exactly the peers arcade can broadcast to — the set `/health` displays.

### Plus: the floor is now explainable

`refreshPolicyOnce` logs which peer set the floor, so "why is the floor 1?" no longer requires a store query.

## Operator note

If a deployment is currently relying on network observation to reach a 0 floor, it will now sit at 1 sat/kB. Set `validator.accept_zero_fee: true` to keep a zero-fee intake — that is what the flag is for, and it is the honest way to express the intent.

Existing `peer_policies` rows for URL-less peers age out of the computation on the normal 15-minute TTL; no migration or manual cleanup is needed.

## Verification

```
go build ./... && go vet ./...          # clean
go test ./... && go test -race ./...    # all pass
go test -tags=postgres ./store/postgres # pass
golangci-lint run                       # 0 issues
```

Both fixes are pinned by tests that **fail against the previous behaviour** — I reverted each fix in turn to confirm:

```
--- FAIL: TestDiscoveredFeePerKB/a_single_zero-fee_peer_cannot_disable_fee_enforcement
--- FAIL: TestRefreshPolicyOnce_ZeroFeePeerDoesNotZeroTheFloor
    intake fee floor = 0, want 1 — a zero-fee peer must not make arcade accept
    transactions no node will mine
--- FAIL: TestPeerPolicyOnlyRecordedForRegisteredURLs/peer_advertising_no_URL_does_not_count
    a peer with no registered URL must not set the network fee, got
    [{PeerID:sender MiningFeeSatoshis:0 MiningFeeBytes:1000 ...}]
```

That last one reproduces the production symptom directly.

Coverage also includes: the cheapest peer still tracked normally, a whole network at zero still flooring, 1 sat/kB passing through unchanged, a stale zero-fee peer ignored entirely, `accept_zero_fee` still pinning 0, and the peer id reported for the log line.

## Caveat

I could not confirm from outside **which** of the two defects produced the live 0 — that needs one query against the deployment's store:

```sql
SELECT peer_id, mining_fee_satoshis, mining_fee_bytes, last_seen
FROM peer_policies WHERE network = 'mainnet'
ORDER BY mining_fee_satoshis ASC LIMIT 5;
```

A fresh row at 0 confirms it. If the lowest is 100, then the live deployments have `accept_zero_fee` set and this PR is a hardening change rather than the fix for that specific symptom — worth checking before assuming the deploy resolves it.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -161,9 +161,16 @@ health:
 #   # announcements — and BOTH enforces it at intake and advertises it via
 #   # GET /policy (issue #212). Until a peer is heard it uses the built-in
 #   # 100 sat/kB floor. Set a non-zero value to pin and enforce an exact floor.
+#   # An observed fee is floored at 1 sat/kB — peer announcements are
+#   # unauthenticated, so one peer advertising 0 must not be able to disable
+#   # fee enforcement for the whole instance. Only accept_zero_fee below can
+#   # produce a 0 floor. Only peers whose datahub URL is registered count
+#   # toward the observation, so /policy's inputs are the endpoints /health
+#   # shows.
 #   min_fee_per_kb: 0
 #   # accept_zero_fee, when true, pins the floor to 0 sat/kB so any tx
-#   # (including fee=0) is accepted at intake, and /policy advertises 0.
+#   # (including fee=0) is accepted at intake, and /policy advertises 0. This
+#   # is the ONLY route to a zero floor; network observation cannot reach it.
 #   # Intended for teratestnet/regtest where the upstream Teranode is also
 #   # configured to accept zero-fee txs. Leave false on mainnet and testnet.
 #   # accept_zero_fee: false
--- a/config/config.go
+++ b/config/config.go
@@ -932,6 +932,11 @@ type ValidatorConfig struct {
 	// fee observed from peer node_status announcements, falling back to
 	// DefaultValidatorMinFeePerKB when no peer has been heard. A non-zero value
 	// (or AcceptZeroFee) is reported verbatim.
+	//
+	// An observed fee is floored at 1 sat/kB: peer announcements are
+	// unauthenticated and the rule is a bare minimum, so a single peer
+	// advertising 0 would otherwise disable fee enforcement entirely. Setting
+	// AcceptZeroFee is the only way to reach a 0 floor.
 	MinFeePerKB uint64 `mapstructure:"min_fee_per_kb"`
 
 	// AcceptZeroFee, when true, pins the validator's fee floor to exactly
--- a/services/api_server/policy_refresher.go
+++ b/services/api_server/policy_refresher.go
@@ -91,11 +91,14 @@ func (s *Server) refreshPolicyOnce(ctx context.Context) {
 	// Fee: the lowest any peer will accept, so arcade rejects only what no peer
 	// would mine. Sizes: the highest any peer will accept, the same rule read
 	// the other way round.
-	fee := uint64(config.DefaultValidatorMinFeePerKB)
+	var (
+		fee      uint64
+		cheapest string // peer that set the floor, for the log line below
+	)
 	if s.feePinned() {
 		fee = s.validator.MinFeePerKB() // pinned at construction; leave it be
-	} else if observed, ok := lowestObservedFeePerKB(peers, ttl, now); ok {
-		fee = observed
+	} else {
+		fee, cheapest = discoveredFeePerKB(peers, ttl, now)
 	}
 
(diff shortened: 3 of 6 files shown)


galt-tr:
## Confirmed against production

Ran the query on `arcade-v2` (namespace `arcade-v2`, `postgres-0`). Cause **(b)** — a live zero-fee peer. `accept_zero_fee` is **not** set anywhere (no validator/fee keys in the deployment env or the `arcade-v2-env` configmap), so this is entirely the peer data.

`peer_policies` — 9 rows, all mainnet, **all last seen within ~10s of each other**:

| fee (sat/kB) | peers |
|---|---|
| **0** | 1 |
| **1** | 2 |
| 100 | 6 |

`datahub_endpoints` — 9 rows:

| policy | last_seen | url |
|---|---|---|
| *(null)* | 2026-07-10 | `http://asset:8090/api/v1` |
| *(null)* | 2026-07-22 | `https://bsva-ovh-teranode-eu-2.bsvb.tech/api/v1` |
| *(null)* | 2026-05-26 | `http://bsva-ovh-teranode-eu-3.bsvb.tech:8000/api/v1` |
| 100 | 2026-08-18 01:13 | the six actively-registering endpoints |

The arithmetic is exact: **six peers at 100 = the six endpoints still registering**. The three anomalous peers (0, 1, 1) are precisely the three whose datahub rows are stale — weeks to months old. Their URLs no longer register, so `UpsertDatahubEndpoint` never runs and `/health` shows them without a policy (or not at all) — but `recordPeerPolicy` runs *ahead of the URL gate* and keeps their fee rows fresh every ~10 seconds.

Note the first row: `http://asset:8090/api/v1`, a teranode announcing its cluster-internal service name — the exact "poisoned registry" case `app/endpoint_source_test.go` already guards on the read side. It has been in the registry since July 10 and is filtered out of `/health` entirely, while still being eligible to set the network fee floor.

### Effect of this PR on that live data

| | resulting floor |
|---|---|
| today | **0** sat/kB |
| floor alone | 1 (the two 1-sat peers are real, fresh observations) |
| URL gate alone | 100 |
| **both (this PR)** | **100 sat/kB** |

So both halves are load-bearing here, and together they produce the value every reachable node actually requires.

### Separate issue, not fixed here

Nothing prunes `datahub_endpoints`: there is no delete path and no TTL sweep, so `http://asset:8090` has sat in the registry for five weeks and `bsva-ovh-teranode-eu-3` since May. After this PR they are harmless to the fee, but they still occupy `/health` as permanently-unhealthy entries. Worth a follow-up.

galt-tr:
Filed the registry-pruning follow-up as #320 so it doesn't get lost with this PR.

PR #321, opened 2026-08-18
fix(policy): a peer's 0 sat/kB is an absent fee, not a free miner

galt-tr:
## What Changed

- `lowestObservedFeePerKB` skips peer rows with `MiningFeeSatoshis == 0`, treating them as "did not advertise a fee" — the same sentinel `highestObservedLimit` already applies to the size limits.
- `minDiscoveredFeePerKB` and its clamp are deleted. With zeros discarded the branch is unreachable (`ceilFeePerKB` returns 0 only for a 0 satoshi input), and its doc comment asserted the reading that caused the bug.
- `isFreshObservation` replaces the inlined freshness check in **both** aggregators. An undated row is now stale rather than immortal.
- `recordPeerPolicy` takes the registered datahub URL and labels the three per-peer gauges with it instead of `msg.BaseURL`; label renamed `base_url` → `datahub_url`.
- `refreshPolicyOnce` logs `fee_unadvertised_peers` alongside `cheapest_peer_id`.
- Docs corrected in `config.example.yaml`, `config/config.go`, `store/store.go` — all three still described the clamp as the design.

## Why It Was Necessary

`arcade-v2-us-1` advertised `miningFee 1 sat/kB` on `GET /policy` while all six healthy endpoints in `GET /health` advertised 100. `maxtxsizepolicy`/`maxscriptsizepolicy` were the discovered values, so discovery was working — only the fee was wrong. ARC clients read `/policy.miningFee` to compute what they pay, so arcade was telling wallets to build transactions at 1/100th of the rate mainnet miners require.

This is the same defect #319 reported, at a different number. That report saw a 0 floor; #319 clamped the observed minimum up to 1. Its own commit message named the cause — *"teranode advertises min_mining_tx_fee=0 when its policy settings are nil"* — which says 0 means **not advertised**. Clamping the result rather than discarding the non-observation moved the symptom from 0 to 1 and left the cause in place.

The production `peer_policies` rows confirm it:

| fee (sat/kB) | rows | last_seen |
|---|---|---|
| 0 | 1 | 02:09:50 |
| 1 | 2 | 02:09:45, 02:09:53 |
| 100 | 6 | live |

The three anomalous rows stop at the exact second the p2p-client rolled from `v0.13.2` to `v0.13.3` (`2026-08-18T02:09:45Z`) — last writes of the pre-#319 binary, which recorded policy ahead of the datahub-URL gate. They aged out after the 15-minute TTL and `/policy` returned to 100 on its own, so this PR is hardening rather than an outage fix. The defect is still live: any peer that *does* register a URL and advertises 0 puts `/policy` back at 1.

## Testing Performed

- `go test ./services/api_server/... ./services/p2p_client/... ./store/... ./validator/... ./config/...` — all pass.
- `go test -race -run 'Policy|Fee|Peer' ./services/api_server/... ./services/p2p_client/...` — pass.
- `golangci-lint run` on all touched packages — 0 issues.
- Every new assertion confirmed **red** before the fix (10 total). `TestRefreshPolicyOnce_MainnetRegression` reproduces the exact production row shape — six peers at 100 plus one nil-policy node — and returned 1 before, 100 now. `TestRecordPeerPolicy_GaugeCoversPropagationOnlyPeers` was verified by temporarily restoring the old label source: it reported `0, want 100`.
- `TestRefreshPolicyOnce_AcceptZeroFeeStillPinsZero` untouched and passing — teratestnet (`accept_zero_fee: true`) must keep advertising 0.
- Pre-existing unrelated failure: `logfields.TestNoCanonicalKeyLiteralsOutsidePackage`, which fails identically on clean `main` (all findings come from a stale local `.claude/worktrees` copy).

## Impact / Risk

Low. Behaviour changes only where a peer advertised no fee, which previously produced a floor no node on the network would honour. A peer genuinely advertising 1 sat/kB is still tracked — pinned by its own test so this cannot silently become a blanket floor.

The `base_url` → `datahub_url` label rename affects three gauges (`arcade_p2p_peer_min_mining_fee`, `..._max_tx_size_policy`, `..._max_script_size_policy`). Nothing in this repo or `bsva-infra-flux` references them; `arcade_p2p_peer_best_height` is genuinely base-URL-keyed and is left alone.

Two follow-ups this PR deliberately does not take on:

1. **Advertised vs enforced fee.** `validator.MinFeePerKB()` still serves both the intake floor and the `/policy` advertisement. MIN-across-peers is right for the first and wrong for the second — and mainnet currently has two peers genuinely advertising 1 sat/kB, so this is concrete rather than hypothetical.
2. **Stale rows are never deleted from `peer_policies`,** only TTL-filtered. The three poisoned rows are still in the production table, inert but re-armable if those peer IDs ever register a URL again.

## Notifications
- @galt-tr

https://claude.ai/code/session_01K8pDVtifV8xhKfhEya8wcJ


raspi-user:
I reviewed PR head `de69274f33f11e1858337f39c77398420086936f`.

No blocking issues found.

This looks like the right correction. Treating `MiningFeeSatoshis == 0` as "not advertised" avoids letting one nil-policy peer drag `/policy.miningFee` down to zero or to the previous clamp. The important cases appear covered: real `1 sat/kB` fees still pass through, `accept_zero_fee` remains the explicit route for intentional zero-fee advertising, undated rows are stale, malformed zero-byte fee rows are skipped, and size limits remain floored at Arcade defaults.

The observability change also looks sound. Recording peer policy only after datahub URL validation keeps `/policy` derived from the same peer set exposed through `/health`, and using `datahub_url` makes propagation-only peers visible in metrics.

Two non-blocking follow-ups I would want tracked:

1. The metric label rename from `base_url` to `datahub_url` may break external dashboards or alerts. If there are no out-of-band consumers, or this is called out in release notes, that is fine.

2. Stale `peer_policies` rows are still TTL-filtered rather than deleted. Acceptable for this fix, but worth keeping as a follow-up so the table does not grow indefinitely.

Checks reviewed: GitHub shows the relevant test/lint/smoke/security checks passing. I did not rerun the full suite locally.

PR #336, opened 2026-09-01
[Propagation/API] Terminal give-up stays terminal; callback registration idempotent per (txid, url, token)

### LINKED ISSUE #335: REJECTED tx with "giving up" extraInfo keeps retrying (retryCount 1,483 and climbing) and fans a status webhook to every accumulated callback registration
We operate an overlay service that broadcasts through arcade-v2-us-1 with
`X-CallbackUrl` registered. We hit a failure mode that looks like two
compounding bugs on the Arcade side, and one clarification request. Our own
side of this (unbounded resubmits stacking up registrations) was our bug and
is fixed as of today; this issue is about what Arcade does with the state we
left behind.

## 1. The durable retry engine does not stop after declaring "giving up"

Example txid (one of about 51 in this state for us):

```
a065a557b3b8afff2033a1de4fe426935d4b6d57d7eec6611f8e83236468e700
```

`GET /tx/{txid}` returns:

```json
{
  "txStatus": "REJECTED",
  "extraInfo": "no network verdict after 288 durable retry attempts: giving up; last network response: PROCESSING (4): [ProcessTransaction][a065a557...] failed to validate transaction",
  "retryCount": 1483,
  "nextRetryAt": "0001-01-01T00:00:00Z",
  "merkleRegisteredAt": "2026-09-01T12:32:57.37148Z"
}
```

The extraInfo says the engine gave up after 288 attempts, but `retryCount`
is 1,483 and increments about once every 10 minutes, days later. The
transactions in question are genuinely invalid (double spends of already
spent UTXOs), so they can never validate; a terminal REJECTED that actually
terminates seems like the intended behavior.

## 2. Status callbacks are delivered once per registration, and registrations accumulate per POST

Each retry cycle flips the status (PENDING_RETRY / REJECTED) and each flip
appears to be delivered to our callback URL once per registration ever made
for that txid. Because our (former) retry ladder re-POSTed these txs many
times, single status flips now arrive at our endpoint thousands of times.
Measured at our callback endpoint: roughly 370 POSTs per second sustained,
about 32 million per day for a week, all for the ~51 txids above, statuses
exclusively REJECTED and PENDING_RETRY. We answer 200 in ~1 ms so delivery
retry is not the multiplier; the registration count is.

Question: is per-(txid, URL) registration deduplication intended? If a
repeat POST of the same txid with the same X-CallbackUrl is supposed to be
idempotent, it does not appear to be.

## Asks

1. A way to cancel or purge callback registrations for a txid set or for a
   callback URL (we can supply the full txid list privately).
2. Make REJECTED-after-giving-up actually terminal in the retry engine.
3. Clarify or implement registration dedup per (txid, callbackUrl).

Happy to provide the full txid list, timing data, or run tests against a
staging endpoint.

Calgooon:
Fixes #335.

## What Changed

Two independent fixes, one commit each.

**1. `fix(propagation)`: a terminal tx never re-enters durable retry; lattice-refused writes are not published.**
- `schedulePendingRetry` now skips a row whose status is already terminal (REJECTED / DOUBLE_SPEND_ATTEMPTED / MINED / IMMUTABLE). New `arcade_propagation_pending_retry_total{outcome="skipped_terminal"}`.
- `BumpRetryCount` leaves a terminal row's `retry_count` untouched (returns the current value) on postgres, pebble and aerospike. `ClearRetryState` is lattice-guarded the same way `SetPendingRetryFields` already was.
- `applyTerminalStatuses` and `persistCascade` publish (and, for cascades, schedule) only rows the lattice actually moved. The dispatcher notification stays unconditional, exactly as the existing comment requires.

**2. `fix(api_server)`: callback registration is idempotent per (txid, callback URL, callback token).**
- The submission id is derived from the subscription identity (`sha256(txid || 0 || url || 0 || token)`, 32 hex chars, the width the random ids used) instead of 16 random bytes.
- `InsertSubmission` is idempotent on that id on all three backends: the latest `X-FullStatusUpdates` intent wins, `last_delivered_status` / retry state / `created_at` are preserved.
- `services/webhook` collapses rows that were duplicated before this change per (callback URL, token) at dispatch, with a stable choice, so historical duplicates deliver once.
- README and OpenAPI document the idempotency; the README's webhook backoff description now matches the code (exponential 5 s to 5 min, not linear).

## Why It Was Necessary

Observed in production (details in #335): a client re-presented ~51 network-invalid transactions through `POST /tx` with `X-CallbackUrl` for a few hours. Each POST appended a new `submissions` row, so those txids ended up with thousands of registrations each. Later, every further POST of a txid Arcade already held as REJECTED (`shouldRepublishExisting` republishes REJECTED rows) ran the fast-path budget, had its PENDING_RETRY write refused by the lattice, and still reached `schedulePendingRetry`: `BumpRetryCount` had no status predicate, so `retry_count` climbed to 1,483 on a row whose `extra_info` already read "giving up", `ClearRetryState` re-wrote the same REJECTED verdict, and a fresh REJECTED event (preceded by a phantom PENDING_RETRY from `applyTerminalStatuses`) was delivered once per accumulated registration. At one subscriber that was ~370 webhook POSTs per second, sustained for days, with `nextRetryAt` reading as the zero time because `ClearRetryState` had nulled it.

The give-up has to be terminal in the scheduler and the stores, not only in the lattice, and a registration has to be a set, not a log.

## Testing Performed

- `CGO_ENABLED=1 go test ./services/propagation/ ./services/webhook/ ./services/api_server/ ./store/pebble/` green locally (the untagged tiers).
- `go vet -tags=postgres ./store/postgres/` green (the embedded-postgres tier was not executed here; the postgres cells mirror the pebble ones line for line, per the existing parity convention).
- New cells: `TestSchedulePendingRetry_TerminalRowIsNeverBumped` (all four terminal statuses: no bump, no clear, no event, `exhausted` unchanged), `TestApplyTerminalStatuses_LatticeRefusedRowIsNotPublished` (refused → silent, real transition → published), `TestBumpRetryCount_TerminalRowUntouched`, `TestClearRetryState_RespectsStatusLattice` (pebble + postgres), `TestHandleSubmitTransaction_RepeatPostSameCallback_RegistersOnce`, `TestSubmissionIDFor_DeterministicAndIdentityBound`, `TestInsertSubmission_SameIDIsIdempotentAndKeepsDeliveryState` (pebble + postgres), `TestDispatchOne_DuplicateRegistrations_DeliverOnce`, `TestCollapseDuplicateRegistrations_PrefersRowWithDeliveryState`.
- Every existing test in those packages still passes.

## Impact / Risk

- Behaviour change for resubmits of a terminal txid: they are still republished to Kafka and re-broadcast per the existing contract, but they no longer touch `retry_count`, no longer re-write the verdict, and no longer emit events. A client polling `GET /tx` sees the same status; it just stops seeing a `retryCount` that climbs forever.
- Behaviour change for repeat registrations: a client that intentionally registered the same URL twice (to get two deliveries) gets one. I could not find a reason anyone would rely on that.
- Historical duplicate rows are not deleted (no schema change, no migration); they are collapsed at dispatch. A follow-up purge would be easy if wanted.
- Aerospike paths are compiled and vetted but not exercised (no cluster here); the changes mirror the pebble logic and the existing `SetPendingRetryFields` guard in that file.
- Not addressed on purpose: whether `shouldRepublishExisting` should keep republishing a REJECTED row at all. That is a contract question for you; with this PR a republish is at least side-effect free on the retry ledger and the event stream.

## Notifications

Happy to split into two PRs, adjust naming, or add an aerospike integration run if you can point me at the harness you prefer.

### DIFF
--- a/services/propagation/propagator.go
+++ b/services/propagation/propagator.go
@@ -578,7 +578,13 @@ func (p *Propagator) applyTerminalStatuses(ctx context.Context, terminalStatuses
 				WithLabelValues(string(prev.Status), string(st.Status)).
 				Observe(time.Since(prev.Timestamp).Seconds())
 		}
-		if prev.Status == st.Status {
+		// A lattice REFUSAL is a no-op too: the store hands back the row it
+		// declined to move (e.g. PENDING_RETRY over REJECTED), and publishing
+		// the requested status would announce a transition that never
+		// happened. That phantom PENDING_RETRY, followed by the give-up path's
+		// re-emitted REJECTED, was the status "flip" a resubmitted terminal tx
+		// broadcast to every registration on every POST.
+		if prev.Status == st.Status || !st.Status.CanTransitionFrom(prev.Status) {
 			continue
 		}
 		switch st.Status {
@@ -883,13 +889,36 @@ func (p *Propagator) persistCascade(ctx context.Context, cascaded []cascadeRejec
 		)
 		p.logger.Warn("transactions parked behind an ancestor with no network verdict", parkFields...)
 	}
-	if _, err := p.store.BatchUpdateStatusReturning(ctx, statuses); err != nil {
+	prevs, err := p.store.BatchUpdateStatusReturning(ctx, statuses)
+	if err != nil {
 		p.logger.Warn(
 			"cascade status write failed",
 			logfields.Status(string(status)),
 			zap.Int("count", len(cascaded)),
 			zap.Error(err),
 		)
+	}
+	// Publish and schedule only the rows the store actually moved. A cascade
+	// can re-visit a descendant that is already terminal (a resubmitted
+	// REJECTED chain, a MINED child of a late-rejected parent); the lattice
+	// declines those writes and hands back the unchanged row, and announcing
+	// the requested status for them would be the same phantom event
+	// applyTerminalStatuses filters. An unknown prev (per-row store fault)
+	// keeps the pre-existing behaviour: publish, so a subscriber is never
+	// starved of an outcome by a transient read error.
+	moved := make([]cascadeRejection, 0, len(cascaded))
+	movedStatuses := make([]*models.TransactionStatus, 0, len(statuses))
+	movedTxids := make([]string, 0, len(txids))
+	for i, cr := range cascaded {
+		if i < len(prevs) && prevs[i] != nil {
+			prev := prevs[i].Status
+			if prev == status || !status.CanTransitionFrom(prev) {
+				continue
+			}
+		}
+		moved = append(moved, cr)
+		movedStatuses = append(movedStatuses, statuses[i])
+		movedTxids = append(movedTxids, cr.txid)
 	}
 	// A cascade-parked descendant has to enter the durable retry queue too.
 	// It reaches PENDING_RETRY without ever being broadcast — its ancestor
@@ -898,7 +927,7 @@ func (p *Propagator) persistCascade(ctx context.Context, cascaded []cascadeRejec
 	// it with a NULL next_retry_at, which GetReadyRetries cannot see: the row
 	// would sit at PENDING_RETRY forever with nothing scheduled to retry it.
 	if status == models.StatusPendingRetry {
-		for _, cr := range cascaded {
+		for _, cr := range moved {
 			// The descendant was never broadcast, so nothing on the network
 			// ever spoke about it; its ancestor's park is the whole story.
 			p.schedulePendingRetry(ctx, cr.txid, cr.rawTx, cascadeReason(status, cr.ancestor))
@@ -908,10 +937,10 @@ func (p *Propagator) persistCascade(ctx context.Context, cascaded []cascadeRejec
 		// statuses (not txids): every cascaded child's reason names ITS OWN
 		// ancestor, so the publish has to group by reason rather than fan one
 		// payload-free event across the whole cascade — see publishRejections.
-		p.publishRejections(ctx, statuses, now)
+		p.publishRejections(ctx, movedStatuses, now)
 		return
 	}
-	p.publishBulkStatus(ctx, status, txids, now)
+	p.publishBulkStatus(ctx, status, movedTxids, now)
 }
 
 // publishBulkStatus fans a post-broadcast batch status update onto the
@@ -2365,6 +2394,21 @@ func giveUpReason(maxAttempts int, lastReason string) string {
 // got a confident explanation of something that never happened.
 func (p *Propagator) schedulePendingRetry(ctx context.Context, txid string, rawTx []byte, lastReason string) {
 	if p.store == nil || len(rawTx) == 0 {
+		return
+	}
+	// A terminal row never re-enters the durable retry queue. Without this
+	// guard a resubmit of a REJECTED txid (handleSubmitTransaction republishes
+	// REJECTED rows by contract) ran the fast-path budget, hit the lattice on
+	// the PENDING_RETRY write, and still landed here: BumpRetryCount had no
+	// status predicate, so retry_count climbed past the budget on every POST
+	// (observed: 1,483 on a tx whose extra_info already read "giving up"),
+	// ClearRetryState re-wrote the same REJECTED verdict, and a fresh REJECTED
+	// event fanned out to every webhook registration each time. The give-up
+	// has to be terminal in the scheduler too, not only in the lattice.
+	if row, err := p.store.GetStatus(ctx, txid); err == nil && row != nil && row.Status.IsTerminal() {
+		metrics.PropagationPendingRetryTotal.WithLabelValues("skipped_terminal").Inc()
+		p.logger.Debug("durable retry not scheduled: tx is already terminal",
+			logfields.TxID(txid), logfields.Status(string(row.Status)))
 		return
 	}
 	count, err := p.store.BumpRetryCount(ctx, txid)
--- a/services/webhook/service.go
+++ b/services/webhook/service.go
@@ -301,7 +301,7 @@ func (s *Service) dispatchOne(ctx context.Context, status *models.TransactionSta
 		return
 	}
 	enriched := false
-	for _, sub := range subs {
+	for _, sub := range collapseDuplicateRegistrations(subs) {
 		if sub.CallbackURL == "" {
 			continue // SSE-only subscription; no webhook to send
 		}
@@ -321,6 +321,44 @@ func (s *Service) dispatchOne(ctx context.Context, status *models.TransactionSta
 		}
 		s.deliver(ctx, sub, status)
 	}
+}
+
+// collapseDuplicateRegistrations keeps ONE submission per (callback URL,
+// callback token) for a txid. Registrations are idempotent at intake now
+// (submissionIDFor), but rows written before that — one per POST, thousands
+// per txid for a client that re-presented a rejected tx for hours — are still
+// stored, and delivering a status once per duplicate row is exactly the
+// webhook flood this guards against. The choice must be STABLE across
+// dispatches so the CAS-maintained delivery state keeps living on one row:
+// prefer a row that already carries delivery state, then the smallest id.
+// Un-chosen rows are never delivered, so they never enter retry state.
+func collapseDuplicateRegistrations(subs []*models.Submission) []*models.Submission {
+	if len(subs) < 2 {
+		return subs
+	}
+	type key struct{ url, token string }
+	chosen := make(map[key]*models.Submission, len(subs))
+	order := make([]key, 0, len(subs))
+	for _, sub := range subs {
+		k := key{sub.CallbackURL, sub.CallbackToken}
+		cur, seen := chosen[k]
+		if !seen {
+			chosen[k] = sub
+			order = append(order, k)
+			continue
+		}
+		curHasState := cur.LastDeliveredStatus != ""
+		subHasState := sub.LastDeliveredStatus != ""
+		if (subHasState && !curHasState) ||
+			(subHasState == curHasState && sub.SubmissionID < cur.SubmissionID) {
+			chosen[k] = sub
+		}
+	}
+	out := make([]*models.Submission, 0, len(order))
+	for _, k := range order {
+		out = append(out, chosen[k])
+	}
+	return out
 }
 
 // shouldDeliver implements the FullStatusUpdates / dedup gating rules:
--- a/services/webhook/service_test.go
+++ b/services/webhook/service_test.go
@@ -854,3 +854,77 @@ func TestReaper_LeaseDenied_SkipsTick(t *testing.T) {
 		t.Fatalf("expected 0 POSTs from non-leader tick, got %d", got)
 	}
 }
+
+// TestDispatchOne_DuplicateRegistrations_DeliverOnce: rows registered before
+// intake became idempotent (one per POST, thousands per txid in the
+// bsv-blockchain/arcade#335 flood) are collapsed per (callback URL, token)
+// at dispatch, so one status event is delivered ONCE to that subscriber.
+// The chosen row is stable (delivery state first, then smallest id), so the
+// CAS-maintained state keeps living on one row across dispatches.
+func TestDispatchOne_DuplicateRegistrations_DeliverOnce(t *testing.T) {
+	var hits atomic.Int32
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		hits.Add(1)
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	st := &fakeStore{
+		subs: map[string][]*models.Submission{
+			txA: {
+				{SubmissionID: "sub-c", TxID: txA, CallbackURL: srv.URL, CallbackToken: "tok"},
+				{SubmissionID: "sub-a", TxID: txA, CallbackURL: srv.URL, CallbackToken: "tok"},
+				{SubmissionID: "sub-b", TxID: txA, CallbackURL: srv.URL, CallbackToken: "tok"},
+				// A different subscriber of the same txid is still its own delivery.
+				{SubmissionID: "sub-z", TxID: txA, CallbackURL: srv.URL, CallbackToken: "other"},
+			},
+		},
+	}
+	svc := New(
+		config.WebhookConfig{HTTPTimeoutMs: 1000, MaxRetries: 3},
+		config.CallbackConfig{AllowPrivateIPs: true},
+		zap.NewNop(), recordingPub{}, st, nil,
+	)
+
+	svc.handleUpdate(t.Context(), &models.TransactionStatus{
+		TxID:      txA,
+		Status:    models.StatusMined,
+		Timestamp: time.Now(),
+	})
+
+	if hits.Load() != 2 {
+		t.Fatalf("expected 2 deliveries (one per distinct subscriber), got %d", hits.Load())
+	}
+	st.mu.Lock()
+	defer st.mu.Unlock()
+	var delivered []string
+	for _, d := range st.deliveries {
+		delivered = append(delivered, d.SubmissionID)
+	}
+	for _, id := range delivered {
+		if id == "sub-b" || id == "sub-c" {
+			t.Errorf("delivery state landed on a non-chosen duplicate row %q (want sub-a, the smallest id); all: %v", id, delivered)
+		}
+	}
+}
+
+func TestCollapseDuplicateRegistrations_PrefersRowWithDeliveryState(t *testing.T) {
+	subs := []*models.Submission{
+		{SubmissionID: "sub-a", CallbackURL: "u", CallbackToken: "t"},
+		{SubmissionID: "sub-b", CallbackURL: "u", CallbackToken: "t", LastDeliveredStatus: models.StatusSeenOnNetwork},
+		{SubmissionID: "sub-c", CallbackURL: "u", CallbackToken: "t"},
+	}
+	got := collapseDuplicateRegistrations(subs)
+	if len(got) != 1 || got[0].SubmissionID != "sub-b" {
+		t.Fatalf("want the row carrying delivery state, got %+v", got)
+	}
+	// Order of first appearance is preserved across distinct subscribers.
+	subs = []*models.Submission{
+		{SubmissionID: "1", CallbackURL: "u2", CallbackToken: "t"},
+		{SubmissionID: "2", CallbackURL: "u1", CallbackToken: "t"},
+	}
+	got = collapseDuplicateRegistrations(subs)
+	if len(got) != 2 || got[0].SubmissionID != "1" || got[1].SubmissionID != "2" {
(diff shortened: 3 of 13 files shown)


services/propagation/propagator.go
```diff
@@ -2367,6 +2396,21 @@ func (p *Propagator) schedulePendingRetry(ctx context.Context, txid string, rawT
...
+	// (observed: 1,483 on a tx whose extra_info already read "giving up"),
+	// ClearRetryState re-wrote the same REJECTED verdict, and a fresh REJECTED
+	// event fanned out to every webhook registration each time. The give-up
+	// has to be terminal in the scheduler too, not only in the lattice.
+	if row, err := p.store.GetStatus(ctx, txid); err == nil && row != nil && row.Status.IsTerminal() {
+		metrics.PropagationPendingRetryTotal.WithLabelValues("skipped_terminal").Inc()
+		p.logger.Debug("durable retry not scheduled: tx is already terminal",
+			logfields.TxID(txid), logfields.Status(string(row.Status)))
+		return
+	}
```
galt-tr: **Important** The guard closes the REJECTED-resubmit flood, but the give-up branch below (L2427-2440) still cannot tell whether `ClearRetryState` applied or was refused: all three backends now return `nil` on a lattice refusal, and the caller falls through to `publishRejections` unconditionally. The window is `GetStatus` -> `BumpRetryCount` -> `ClearRetryState`; a MINED/IMMUTABLE write landing in it (`bump_builder` -> `SetMinedByTxIDs` can hit any in-flight row, including PENDING_RETRY) leaves the store at MINED while a REJECTED event goes out to every webhook/SSE subscriber. This is the exact scenario the new `TestClearRetryState_RespectsStatusLattice` cell exercises at the store level (a MINED row asked to become REJECTED by a stale give-up stays MINED); before this PR the store was clobbered too, so this is strictly an improvement, but the event/store divergence remains.

Suggested shape, mirroring `UpdateDeliveryStatusCAS`:

```go
// store.Store
ClearRetryState(ctx, txid, finalStatus, extraInfo) (applied bool, err error)
```
Postgres already has `tag.RowsAffected()` from the `Exec`; pebble/aerospike return `false` on the guard path. Then in the caller: `if !applied { return }` before `publishRejections`.

Related: folding the terminal check into `BumpRetryCount` (e.g. return a `store.ErrTerminal` sentinel, or `(count, status)`) would remove both the TOCTOU window and the extra `GetStatus` round-trip this adds to every call site, including the reaper paths (`reaper.go:429`, `:571`) where the row was selected as PENDING_RETRY moments earlier. Note `GetStatus` runs `enrichMerklePath`, so for a MINED row the status peek also performs a BUMP lookup.

services/propagation/propagator.go
```diff
@@ -2367,6 +2396,21 @@ func (p *Propagator) schedulePendingRetry(ctx context.Context, txid string, rawT
...
+	// guard a resubmit of a REJECTED txid (handleSubmitTransaction republishes
+	// REJECTED rows by contract) ran the fast-path budget, hit the lattice on
+	// the PENDING_RETRY write, and still landed here: BumpRetryCount had no
+	// status predicate, so retry_count climbed past the budget on every POST
+	// (observed: 1,483 on a tx whose extra_info already read "giving up"),
+	// ClearRetryState re-wrote the same REJECTED verdict, and a fresh REJECTED
+	// event fanned out to every webhook registration each time. The give-up
+	// has to be terminal in the scheduler too, not only in the lattice.
+	if row, err := p.store.GetStatus(ctx, txid); err == nil && row != nil && row.Status.IsTerminal() {
+		metrics.PropagationPendingRetryTotal.WithLabelValues("skipped_terminal").Inc()
```
galt-tr: **Important** (alerting) `skipped_terminal` is a new label value on `arcade_propagation_pending_retry_total` but `metrics/preregister.go:84-86` still instantiates only `scheduled` and `exhausted`. Per the comment on that function, an alert on a series that was never born returns no-data rather than 0 - and this series is the one signal for the re-POST flood this PR fixes, so it is the one operators will want to alert on. Add `"skipped_terminal"` to the slice in `PreRegisterPropagationRetryOutcomes`, and update the trailing comment at `metrics/metrics.go:310` (`// scheduled, exhausted`) plus the doc comment above it so the new outcome is described alongside the other two.

services/propagation/propagator.go
```diff
@@ -883,22 +889,45 @@ func (p *Propagator) persistCascade(ctx context.Context, cascaded []cascadeRejec
...
+		if i < len(prevs) && prevs[i] != nil {
+			prev := prevs[i].Status
+			if prev == status || !status.CanTransitionFrom(prev) {
+				continue
+			}
+		}
+		moved = append(moved, cr)
+		movedStatuses = append(movedStatuses, statuses[i])
+		movedTxids = append(movedTxids, cr.txid)
+	}
```
galt-tr: **Nit** `moved` is only consumed by the `status == models.StatusPendingRetry` branch, while `movedStatuses`/`movedTxids` serve the other two exits. Either build `moved` lazily inside that branch, or drop it and iterate `movedStatuses` there (each entry carries `TxID`; `rawTx` and `ancestor` would need a parallel index, so the simplest is to keep a single `movedIdx []int` and derive all three views from it). Cosmetic; the filter itself (`prevs[i]` <-> `cascaded[i]`, fail-open on a nil prev) reads correctly.

store/pebble/pebble.go
```diff
@@ -1913,6 +1924,22 @@ func (s *Store) InsertSubmission(ctx context.Context, sub *models.Submission) er
...
+		var existing storedSubmission
+		decodeErr := json.Unmarshal(existingRaw, &existing)
+		_ = closer.Close()
+		if decodeErr == nil {
+			existing.FullStatusUpdates = sub.FullStatusUpdates
+			stored = existing
+		}
+	} else if !errors.Is(err, pebbledb.ErrNotFound) {
+		return err
+	}
```
galt-tr: **Important** This read-modify-write is not serialized with `UpdateDeliveryStatusCAS` (L2173) and `RecordDeliveryAttempt` (L2135), both of which take `s.submissionMu` for exactly this reason (see the field doc at L56-60). Now that the id collides by design, every re-registration is a concurrent writer to a row the webhook service is CAS-advancing, from 8 recorder workers. Interleaving Get(here) -> CAS Set -> Set(here) writes the pre-CAS `LastDeliveredStatus` back, so the CAS reported `claimed=true` but the row no longer says so; the next event for that status is then delivered again - the re-delivery this PR is preventing - and a retry-state write from `recordFailure` can be lost the same way.

Reproduced with a throwaway probe (not committed) that races `InsertSubmission` against `UpdateDeliveryStatusCAS` on one id under `-race`: 2150 of 4000 claimed CAS advances read back the stale status on this branch; 0 of 4000 with the lock taken.

```go
func (s *Store) InsertSubmission(ctx context.Context, sub *models.Submission) error {
	if err := ctx.Err(); err != nil {
		return err
	}
	s.submissionMu.Lock()
	defer s.submissionMu.Unlock()
	// ...
```
Also widen the `submissionMu` doc comment (L56-59) to name `InsertSubmission`. Postgres is fine: the upsert is a single atomic statement.

store/pebble/pebble.go
```diff
@@ -1079,6 +1084,12 @@ func (s *Store) ClearRetryState(ctx context.Context, txid string, finalStatus mo
 	if existing == nil {
 		return nil
 	}
+	// Lattice guard (mirrors SetPendingRetryFields): a row whose current
+	// status forbids finalStatus is left untouched. Silent skip matches the
+	// guarded status update's semantics.
+	if !finalStatus.CanTransitionFrom(models.Status(existing.Status)) {
+		return nil
+	}
```
galt-tr: **Minor** (companion to the propagator comment) A refused clear and an applied clear are indistinguishable to the caller. Returning `(false, nil)` here (and from the aerospike guard and the postgres `RowsAffected()==0` case) lets `schedulePendingRetry` skip the REJECTED publish when the lattice kept the row. The existing test `TestClearRetryState_RespectsStatusLattice` would then assert the returned flag as well.

store/aerospike/aerospike.go
```diff
@@ -931,11 +931,28 @@ func (s *Store) BumpRetryCount(ctx context.Context, txid string) (int, error) {
...
+		return 0, fmt.Errorf("bump retry count %s: %w", txid, store.ErrNotFound)
+	}
+	if models.Status(getString(rec, "status")).IsTerminal() {
+		if v, ok := rec.Bins["retry_count"]; ok {
+			if n, ok := v.(int); ok {
+				return n, nil
+			}
+		}
+		return 0, nil
+	}
```
galt-tr: **Minor** Two things on this hunk.

1. This branch changes a contract: on a missing record the base `Operate(AddOp)` write-created a record (default `RecordExistsAction` is `UPDATE`), whereas this now returns `store.ErrNotFound` like pebble/postgres do. That is the right parity, but it is undocumented; add a sentence to the method doc comment so the next reader does not treat it as a regression.

2. This backend got all three production changes (here, `ClearRetryState`, `InsertSubmission`) with no test at all: `store/aerospike/*_test.go` has no reference to `BumpRetryCount`, `ClearRetryState` or `InsertSubmission`, and `store/storetest/` covers tracker rows only. `go vet -tags=integration` passes, but a compile is not evidence that the `getString(rec, "status")` / `rec.Bins["retry_count"].(int)` shapes match what the cluster returns. Please add integration-tagged cells mirroring the pebble ones (`TestBumpRetryCount_TerminalRowUntouched`, `TestClearRetryState_RespectsStatusLattice`, `TestInsertSubmission_SameIDIsIdempotentAndKeepsDeliveryState`) so a maintainer with a cluster can run them before this ships to an aerospike deployment.

store/aerospike/aerospike.go
```diff
@@ -1955,6 +1985,17 @@ func (s *Store) InsertSubmission(ctx context.Context, sub *models.Submission) er
...
+	// identity): an existing record keeps its delivery bins and created_at;
+	// only the caller's X-FullStatusUpdates intent is refreshed. A blind Put
+	// would drop last_delivered and re-deliver every status already sent.
+	if rec, gerr := s.client.Get(s.readPolicy(ctx), key, "submission_id"); gerr != nil && !isKeyNotFound(gerr) {
+		return fmt.Errorf("read submission %s: %w", sub.SubmissionID, gerr)
+	} else if rec != nil {
+		_, err := s.client.Operate(s.writePolicy(ctx), key,
+			aero.PutOp(aero.NewBin("full_updates", sub.FullStatusUpdates)))
+		return err
+	}
```
galt-tr: **Minor** The stale-clobber race pebble has does not apply here: `Operate` on the single `full_updates` bin is atomic per record and never touches `last_status`. The interaction is the other way round: every re-registration bumps the record generation, and `UpdateDeliveryStatusCAS` (L2241) uses `EXPECT_GEN_EQUAL`, so a CAS that lands between a registration's Get and its Operate returns `claimed=false` and the replica skips the POST; the webhook reaper will not pick that row up because `retry_count` stays 0. Before this PR a re-POST created a fresh record and never touched the existing one's generation, so this is new exposure under exactly the re-POST-burst shape from #335. Options: have `UpdateDeliveryStatusCAS` re-read and retry once on a generation error (the value it compares is `last_status`, not the generation), or drop the Get here and write with `RecordExistsAction = UPDATE_ONLY` falling back to `CREATE_ONLY`, which at least removes the check-then-act.

store/postgres/postgres.go
```diff
@@ -873,11 +873,26 @@ func (s *Store) GetTxIDsByBlockHash(ctx context.Context, blockHash string) ([]st
 
 // BumpRetryCount is a single-statement atomic increment — no client-side
 // mutex needed because Postgres handles the CAS.
+//
+// A terminal row (REJECTED, DOUBLE_SPEND_ATTEMPTED, MINED, IMMUTABLE) is left
+// untouched and its current count is returned: retry_count is the durable
+// retry budget, and a tx that has already reached its verdict has no budget
+// to spend. Without this predicate every resubmit of a given-up txid kept
+// incrementing the counter past the budget (1,483 on a "giving up" row).
```
galt-tr: **Nit** The two lines above this paragraph (L874-875) still say "single-statement atomic increment - no client-side mutex needed because Postgres handles the CAS". The terminal and missing-row paths are now two statements (guarded UPDATE, then SELECT); the increment itself is still atomic, so reword to e.g. "the increment is a single guarded UPDATE; a terminal or missing row falls back to a plain SELECT of the current count".

services/webhook/service.go
```diff
@@ -323,6 +323,44 @@ func (s *Service) dispatchOne(ctx context.Context, status *models.TransactionSta
...
 
+// collapseDuplicateRegistrations keeps ONE submission per (callback URL,
+// callback token) for a txid. Registrations are idempotent at intake now
+// (submissionIDFor), but rows written before that — one per POST, thousands
+// per txid for a client that re-presented a rejected tx for hours — are still
+// stored, and delivering a status once per duplicate row is exactly the
+// webhook flood this guards against. The choice must be STABLE across
+// dispatches so the CAS-maintained delivery state keeps living on one row:
+// prefer a row that already carries delivery state, then the smallest id.
+// Un-chosen rows are never delivered, so they never enter retry state.
```
galt-tr: **Minor** "Un-chosen rows are never delivered, so they never enter retry state" is true only for rows that were never delivered before this change. Legacy duplicates that already carry `NextRetryAt`/`RetryCount > 0` (likely after the #335 flood, where every row was being POSTed) are still returned by `ListSubmissionsReadyForRetry` and redelivered one row at a time by `reapOnce` (`services/webhook/reaper.go:102-132`), which does not go through `collapseDuplicateRegistrations`. Bounded (each row until success or `max_retries`) and one-time, but worth either applying the collapse in `reapOnce` as well or stating the carve-out here.

services/webhook/service.go
```diff
@@ -323,6 +323,44 @@ func (s *Service) dispatchOne(ctx context.Context, status *models.TransactionSta
...
+		if !seen {
+			chosen[k] = sub
+			order = append(order, k)
+			continue
+		}
+		curHasState := cur.LastDeliveredStatus != ""
+		subHasState := sub.LastDeliveredStatus != ""
+		if (subHasState && !curHasState) ||
+			(subHasState == curHasState && sub.SubmissionID < cur.SubmissionID) {
+			chosen[k] = sub
```
galt-tr: **Minor** When several duplicates carry delivery state (the common post-incident shape, since every row was being delivered), "smallest id" can pick a row whose `LastDeliveredStatus` is behind another's, e.g. `sub-a`=SEEN_ON_NETWORK, `sub-b`=MINED: the next MINED event is delivered again through `sub-a`. One extra POST per (url, token), once, then the chosen row is caught up - so not a flood, but avoidable. Preferring the most advanced status (index in `models.AllStatuses()`) before falling back to the smallest id keeps the choice stable and never re-sends. The tests only cover the one-row-with-state case; a cell with two stateful rows would pin whichever rule is chosen.

services/webhook/service_test.go
```diff
@@ -854,3 +854,77 @@ func TestReaper_LeaseDenied_SkipsTick(t *testing.T) {
...
+		Status:    models.StatusMined,
+		Timestamp: time.Now(),
+	})
+
+	if hits.Load() != 2 {
+		t.Fatalf("expected 2 deliveries (one per distinct subscriber), got %d", hits.Load())
+	}
+	st.mu.Lock()
+	defer st.mu.Unlock()
+	var delivered []string
```
galt-tr: **Nit** CI lint fails on this line (the only finding): `prealloc: Consider preallocating delivered with capacity len(st.deliveries)`. `delivered := make([]string, 0, len(st.deliveries))` clears it. The Test Suite job is skipped for fork PRs, so lint is the only CI signal on this PR right now.

galt-tr (changes requested):
**#335 validity:** retry-engine half **valid** (base `schedulePendingRetry` bumps/clears/publishes with no status predicate; `GetReadyRetries` selects only PENDING_RETRY on all three backends, so the driver is the REJECTED republish on POST, one increment per POST); registration-dedup half **valid** (random 16-byte submission id, `ON CONFLICT DO NOTHING` never fires, `dispatchOne` delivers per row). "REJECTED after give-up is terminal" matches `models.terminalStatuses`.

**Overall.** The mechanism analysis on the issue is correct and the fix lands in the right layers: the scheduler short-circuits terminal rows, `BumpRetryCount`/`ClearRetryState` gain the same lattice discipline `SetPendingRetryFields` already had, `applyTerminalStatuses`/`persistCascade` stop announcing transitions the store refused, and registration becomes a set keyed on `(txid, url, token)` with delivery state preserved on all three backends. Test coverage for pebble and postgres is thorough and both tiers pass locally. Two Important findings remain: pebble `InsertSubmission` is a read-modify-write on a colliding key that does not take `submissionMu` (reproduced: 2150/4000 CAS advances lost under a race probe, 0 with the lock), and `ClearRetryState`'s new silent refusal leaves the give-up path publishing REJECTED for a row the lattice kept at MINED/IMMUTABLE. Both are small, local fixes.

**Findings outside the diff (no inline anchor possible):**
- `metrics/preregister.go:84-86` - `skipped_terminal` is not pre-registered (only `scheduled`, `exhausted`); the function's own comment explains why that matters for alerting. `metrics/metrics.go:310` trailing comment and the doc block above it still list only the two old outcomes. (Also flagged inline on the `.Inc()` line.)
- `services/webhook/reaper.go:102-132` - `reapOnce` redelivers per `ListSubmissionsReadyForRetry` row without `collapseDuplicateRegistrations`; legacy duplicates already in retry state are swept individually (bounded, one-time). See the inline note on the collapse doc comment.
- Commit hygiene: the three backends' `InsertSubmission` idempotency hunks (`store/pebble/pebble.go`, `store/postgres/postgres.go`, `store/aerospike/aerospike.go`) are in `34e97d2` (the propagation commit) although only `submissionIDFor` in `42bd699` needs them (`git show 34e97d2 --stat`). Moving those hunks into `42bd699`, or splitting into two PRs as offered, would make each commit revertable on its own.
- Contract note, not a defect: the identity is `(txid, url, token)` while the issue asked about `(txid, callbackUrl)`. A client that rotates `X-CallbackToken` and re-POSTs gets a second row and a second delivery to the same URL. Defensible (the token is the bearer credential), but worth a sentence in the README so nobody reads "same headers" loosely.

**Verification.** Ran locally at `42bd699` with `CGO_ENABLED=1`: `gofmt -l services/ store/` (clean), `go build ./...`, `go vet` on the four untagged packages plus `-tags=postgres ./store/postgres/` and `-tags=integration ./store/aerospike/` (all clean), `go test ./services/propagation/... ./services/webhook/... ./services/api_server/... ./store/pebble/...` (ok), `go test -tags=postgres ./store/postgres/...` (ok, embedded postgres), `go test -race ./store/pebble/ -run 'Submission|RetryCount|ClearRetryState'` (ok), and `golangci-lint cache clean && golangci-lint run` on the touched trees (1 finding: the prealloc nit at `services/webhook/service_test.go:900`). Not run: the aerospike integration tier (no cluster); those hunks are compile/vet-only here as well.

